Issue #24, opened 2016-06-24
Form focus in modal should be in the first field by default


FRosner:
### Problem

When creating a new instance, the focus is not inside the textbox.
### Solution

Put the focus inside the text box.

FRosner:
Thanks @pliguori

PR #59, opened 2016-08-23
#24: Modal popup - focus on first input field

pliguori:
done by creating a custom attribute directive that runs a conditional focus


FRosner:
Thanks for the contribution, @pliguori. Is it still possible to tab through the fields? Is there a way to solve this more elegant with HTML5?

pliguori:
@FRosner what do you actually mean with "more elegant with HTML5"? If you create the popup via angularjs, you need to manage its behavior via angularjs, else it could become unpredictable.

And of course you can tab through the fields, that's normal behavior based on the DOM (TAB goes to the next enabled input field in the hierarchy)

PR #143, opened 2016-11-10
Confirmation before instance deletion

### LINKED ISSUE #93: Confirmation before instance deletion
### Problem

The end user is not asked for confirmation before deletion of an instance under a template.
### Solution
- [ ] When deleting an instance there should be a modal asking for confirmation.
- [ ] It could also show the parameter values etc. to allow the user to double check.
- [ ] It should also inform about the fact that deletion will also result in submitting a stop request to nomad.

IshmeetKaur:
Modal asking for confirmation when deleting an instance.

fixes #93

### DIFF
--- a/public/newInstanceModal.js
+++ b/public/newInstanceModal.js
@@ -1,5 +1,5 @@
 angular.module('broccoli')
-  .controller('NewInstanceCtrl', function ($scope, $rootScope, $uibModalInstance, template, instance, templates) {
+  .controller('NewInstanceCtrl', function ($scope, $rootScope, $uibModalInstance, template, instance, templates, deleteInstance) {
     var vm = this;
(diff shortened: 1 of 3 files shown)


public/newInstanceModal.js
```diff
@@ -47,11 +56,11 @@ angular.module('broccoli')
     vm.cancel = cancel;
 
     function ok() {
-      Object.keys(vm.paramsToValue).forEach(function(param) {
+        Object.keys(vm.paramsToValue).forEach(function(param) {
```
FRosner, line 59: Let's not change the formatting in the same commit :)

I recommend you look at your changes before you commit them @IshmeetKaur (e.g. with a GUI or `git add -p`).

PR #242, opened 2017-07-14
#240 remove replaceAll and regex as we don't need it and it breaks

### LINKED ISSUE #240: Inserting parameter values fails because of replaceAll
image

```
Failed to update instance: "Invalid request to update an instance: java.lang.IllegalArgumentException: Illegal group reference: group index is missing"
```

Probably related to: https://github.com/FRosner/cluster-broccoli/blob/74e56050add6ceb760c9937145433c173a6cec1a/app/de/frosner/broccoli/models/Instance.scala#L53

FRosner:
Fixes #240


Gerrrr (approved):
LGTM

PR #248, opened 2017-07-17
WIP: Show tasks and allocations for instances

swsnr:
Corresponding project: https://github.com/FRosner/cluster-broccoli/projects/2


server/src/main/scala/de/frosner/broccoli/nomad/NomadConfiguration.scala
```diff
@@ -0,0 +1,21 @@
+package de.frosner.broccoli.nomad
+
+import com.typesafe.config.Config
+
+/**
+  * The Nomad configuration.
+  *
+  * @param url The URL to connect to to access nomad via HTTP.
+  */
+final case class NomadConfiguration(url: String)
```
FRosner: What happens here if the url is not set?

swsnr: In application.conf?  It'll fallback to the one from reference.conf.

FRosner: What will happen to the application if neither application.conf nor reference.conf contains this value?

swsnr: It'll throw an exception.  reference.conf is supposed to contain all configuration settings you use throughout the application; hence the name.

swsnr: PS: reference.conf is not for the user of an application: You shouldn't modify reference.conf during deployment; it's part of the application code essentially

FRosner:
Is the reference.conf packaged inside the jar or the dist zip automatically?

Does play exit cleanly if there is an exception at this point? Will it respect shutdown hooks?

Thanks!

swsnr:
@FRosner  Why should `reference.conf` not be packaged?  It's a resource like any other, isn't it?

I think that Play runs stop hooks no matter how the application exits, but I would not bother to find out in detail.  

As I said `reference.conf` should contain **all** settings of the application; if an exception gets thrown here, that's a programmer error.

swsnr:
@FRosner This is how it looks like with the latest push:

bildschirmfoto 2017-08-02 um 17 28 23

swsnr:
@FRosner I'll close this PR and split it into smaller ones for easier review.

Issue #44, opened 2016-07-20
Show logs of nomad job


FRosner:
Progress on https://github.com/FRosner/cluster-broccoli/projects/2

### Problem

When a nomad job is started through the UI, it would be nice to be able to get logs about this job. A Broccoli instance can create one (one-time batch or system jobs) or multiple (periodic jobs) Nomad jobs. A Nomad job has one or multiple allocations which can be pending, running, failing or finishing. The UI should reflect this in a consistent but understandable way.

### Solution

The Broccoli HTTP API should have a possibility to query for out and err logs of an instance and the UI should show it somewhere.
### References
- https://www.nomadproject.io/docs/http/client-fs.html

swsnr:
# TODO:

- [x] Find out how to get access to the client that holds an allocation, as `client/fs` requests must be made to the client that runs an allocation, see <https://www.nomadproject.io/docs/http/client-fs.html>. Looks as if we can just get the HTTP address of the corresponding node.

# How to get hold of logs

```console
# Get tasks of a job
$ http get localhost:4646/v1/job/test | jq '.TaskGroups | map(.Tasks) | flatten | map(.Name)'
[
  "http-task"
]
# Get allocations of a job
http get localhost:4646/v1/job/test/allocations | jq 'map({ID: .ID, NodeID: .NodeID})'
[
  {
    "ID": "699a84f9-1829-e6cd-d677-1ed10d8c9253",
    "NodeID": "fb2170a8-257d-3c64-b14d-bc06cc94e34c"
  }
]
# Get HTTP address of a node of an allocation
$ http get localhost:4646/v1/node/fb2170a8-257d-3c64-b14d-bc06cc94e34c | jq '.HTTPAddr'
"127.0.0.1:4646"
# Get logs for a task in an allocation (from the node HTTP address as received above)
$ http get "localhost:4646/v1/client/fs/logs/699a84f9-1829-e6cd-d677-1ed10d8c9253?follow=0&task=http-task&type=stderr" | jq -r .Data | base64 -D
```

We probably need to show all tasks & their allocations in the UI, and then provide an option to get logs per each combination of task & allocation.

swsnr:
@FRosner Should be fixed, shouldn't it?  If not, can you please un-assign me?

PR #250, opened 2017-07-18
Use default SBT layout

swsnr:
Move from the Play framework project layout to the [standard SBT layout](https://playframework.com/documentation/2.5.x/Anatomy#Default-SBT-layout).

# What's changed?

* The entire backend code is now in a separate `server` module in the corresponding directory, using the standard Scala layout.  Hence sources are now in `src/main/scala`, resources in `src/main/resources`, assets in `src/main/public`, tests in `src/test/scala`, and so on.
* To run the server, use `sbt server/run` now.  `sbt run` won't work anymore.

# Cons

Apparently the layout is still "experimental", but I've been using it in all my Play apps so far and never had any issues.

# Pros

* It's more in line with standard Scala conventions.
* It enables proper SBT integration tests (`sbt it:test` for tests in `src/it/scala`) which I intend to use for the new nomad service written as part of #44.  Later on we can port the existing HTTP API tests to proper integration tests as well.
* It provides the infrastructure to turn the ELM part into a proper SBT sub-project at some point, and control the ELM build through SBT, so that `sbt compile` and `sbt dist` would ideally build the ELM part too, reducing the overall complexity of our build.


FRosner (approved):
LGTM

Keep in mind that the location of the `public` folder changed to `server/src/main/public`

PR #253, opened 2017-07-20
#251 simplify the account parsing and remove the default instance regex

### LINKED ISSUE #251: Default regex pattern is .*
In our configuration I found a typo:

```
  {username:"abc-admin", password:"pass1", instanceRegexp="^abc-.*", role="administrator" },
  {username:"abc-operator", password:"pass2", instanceRegexp="^abc-.*", role="operator" }
```

Pay attention to `instanceRegexp` instead of the valid `instanceRegex`. This parameter was not used and `instanceRegex` was set to `.*`. Such default can be dangerous.  Should we make broccoli complain when `instanceRegex` is not defined for a user?

Gerrrr:
Fixes #251


swsnr (approved):
Perfect

swsnr:
Incredible 😑

Issue #229, opened 2017-06-22
Reload templates without restarting


mastaofdisasta:
Cluster broccoli could watch for changes in template directory (inotify?) or provide some HTTP API endpoint to request template refresh.

FRosner:
Thanks for creating that issue @mastaofdisasta! I think a good first step is to support template reloading through a signal in case people are deploying the Broccoli templates through the template stanza from Nomad. What do you think?

We could use `SIGUSR1` to reload the templates?

Providing an HTTP endpoint to reload is also possible but we would have to add some authorization around that. I don't see a huge benefit over signals, esp. as reloading only makes sense after changes to the files have been made. Instead, we should allow changing templates through the API / Web UI so that administrators can create new templates during runtime without access to the file system.

PS: Sorry for the late response I was not in the country.

swsnr:
I suggest tearing the `TemplateService` apart and move template loading into the following structure of "layered template sources":

```scala

trait TemplatesSource {
  def loadTemplates(): Seq[Template]
}

class DirectoryTemplatesSources(directory: String) extends TemplatesSource {
  def loadTemplates(): Seq[Template] = ???
}

@Singleton
class CachedTemplatesSource(source: TemplatesSource) extends TemplatesSource {
  @volatile private var templatesCache: Option[Seq[Template]] = ???

  override def loadTemplates(): Seq[Template] =
    templatesCache match {
      case Some(templates) => templates
      case None => {
        refresh()
        templatesCache.get
      }
    }

  def refresh(): Unit = templatesCache = Some(source.loadTemplates())
}

class CacheRefresher(source: CachedTemplatesSource) extends TemplatesSource {

  def wheneverUSR1(): Unit = {
    source.refresh()
  }

  override def loadTemplates(): Seq[Template] = source.loadTemplates()
}
```

Each part represents a single concern now:

1. Loading templates from disk
2. Maintaining an in-memory cache of templates
3. Refreshing the cache on SIGURS1

The TemplateService would get a TemplatesSource injected via DI; a new "TemplatesModule" would configure the appropriate TemplatesSource for injection. /cc @Gerrrr

FRosner:
Is everything done here @Gerrrr ?

Gerrrr:
Not yet.

[...]

PR #254, opened 2017-07-21
Refactor

swsnr:
Some refactorings that both me and @Gerrrr need to work on our issues, so I extracted it into a separate branch.

Should be merged before #248 and @Gerrrr's upcoming PR for #229


FRosner (approved):
LGTM! Thanks a lot :)

PR #255, opened 2017-07-24
Refresh templates by sending SIGUSR2

Gerrrr:
Solves #229

### DIFF
--- a/server/src/main/resources/reference.conf
+++ b/server/src/main/resources/reference.conf
@@ -1,4 +1,4 @@
-# Nomad configuration
 broccoli {
   nomad.url = "http://localhost:4646"
+  templates.path = "templates"
 }
--- a/server/src/main/resources/reference.conf
+++ b/server/src/main/resources/reference.conf
@@ -1,4 +1,4 @@
 broccoli {
   nomad.url = "http://localhost:4646"
-  templates.path = "templates"
+  templates.path = "./templates"
 }
--- a/server/src/main/scala/de/frosner/broccoli/controllers/TemplateController.scala
+++ b/server/src/main/scala/de/frosner/broccoli/controllers/TemplateController.scala
@@ -12,8 +12,6 @@ case class TemplateController @Inject()(templateService: TemplateService, overri
     extends Controller
     with BroccoliSimpleAuthorization {
 
-  private val templates = templateService.getTemplates
-
   def list: Action[AnyContent] = StackAction { implicit request =>
     Ok(Json.toJson(TemplateController.list(templateService)))
   }
--- a/server/src/main/scala/de/frosner/broccoli/services/TemplateService.scala
+++ b/server/src/main/scala/de/frosner/broccoli/services/TemplateService.scala
@@ -3,43 +3,13 @@ package de.frosner.broccoli.services
 import javax.inject.Inject
 import javax.inject.Singleton
 
-import de.frosner.broccoli.conf
 import de.frosner.broccoli.models.Template
 import de.frosner.broccoli.util.Logging
 import de.frosner.broccoli.templates._
-import play.api.Configuration
 
 @Singleton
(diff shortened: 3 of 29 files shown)


server/src/main/scala/de/frosner/broccoli/services/TemplateService.scala
```diff
@@ -37,89 +32,15 @@ class TemplateService @Inject()(configuration: Configuration) extends Logging {
...
-    } else {
+  def getTemplates: Seq[Template] = {
+    if (templatesStorageType != conf.TEMPLATES_STORAGE_TYPE_FILESYSTEM) {
       throw new IllegalStateException(s"$templatesStorageType not supported")
     }
-  }
 
-  def getTemplates: Seq[Template] = templates
+    new SignalRefreshedTemplateSource(new CachedTemplateSource(new DirectoryTemplateSource(templatesUrl)))
+      .loadTemplates()
```
swsnr: Please inject this as dependency.

server/src/main/scala/de/frosner/broccoli/controllers/TemplateController.scala
```diff
@@ -12,8 +12,6 @@ case class TemplateController @Inject()(templateService: TemplateService, overri
     extends Controller
     with BroccoliSimpleAuthorization {
 
-  private val templates = templateService.getTemplates
```
swsnr, line 15: Was this unused?

swsnr (changes requested):
Please add API documentation to all template sources and model case classes.

server/src/main/scala/de/frosner/broccoli/controllers/TemplateController.scala
Gerrrr: yes

swsnr: Okay

swsnr (changes requested):
Please do not forget to add API documentation for the template source trait and implementations :blush:

server/src/main/scala/de/frosner/broccoli/templates/SignalRefreshedTemplateSource.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import de.frosner.broccoli.models.Template
+import sun.misc.{Signal, SignalHandler}
+
+/**
+  * The template source that wraps CachedTemplateSource and refreshes the cache after receiving SIGUSR2
+  *
+  * @param source The CachedTemplateSource that will be wrapped
+  */
+@Singleton
```
swsnr: Please remove this annotation; it's not used since the source is created through a provider method which has its own `@Singleton` annotation.

server/src/main/scala/de/frosner/broccoli/templates/CachedTemplateSource.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import de.frosner.broccoli.models.Template
+import javax.inject.Singleton
+
+/**
+  * The template source that wraps another template source and caches loaded templates of that source
+  *
+  * @param source The template source that will be cached
+  */
+@Singleton
```
swsnr: Please remove this annotation; it's not used since the source is created through a provider method which has its own `@Singleton` annotation.

server/src/main/resources/reference.conf
```diff
@@ -1,4 +1,4 @@
-# Nomad configuration
 broccoli {
   nomad.url = "http://localhost:4646"
+  templates.path = "templates"
```
swsnr: Should we perhaps use `./templates` to make clear that it's a path relative to the local directory?

swsnr (changes requested):

swsnr (approved):
Thanks am happy now! ☺️

PR #258, opened 2017-07-26
Build the frontend through sbt

swsnr:
The frontend is a proper SBT project now; `sbt dist` now builds the webpack bundle via dependencies and resource generators.  The special directory for generated resources is gone; the SBT project copies the generated files into the generated resources directory of SBT and thus hides them from the project tree.


FRosner (dismissed):
LGTM

swsnr:
@FRosner I've updated the branch, can you review again?

FRosner (approved):
lgtm

PR #259, opened 2017-07-27
Prevent npm publish in webui

swsnr:
Builds on #258 (because renaming), and simply sets `private: true` for the NPM package to prevent an accidental `npm publish` actually publishing to the NPM registry.


swsnr:
@FRosner Could you review and merge this one?

FRosner (dismissed):
wow

FRosner (approved):
is ok

PR #260, opened 2017-07-27
Refactor signal handling into separate module

Gerrrr:


server/src/main/scala/de/frosner/broccoli/signal/SignalManager.scala
```diff
@@ -0,0 +1,22 @@
...
+/**
+  * Provide a way to register and unregister OS signals
+  */
+trait SignalManager {
+
+  /**
+    * Registers signal handler for the given signal
+    * @param signal
+    * @param handler
+    */
```
swsnr: The implementation doesn't allow to register the same signal twice, so please document this behaviour.  Ideally, add `@throws` fields below `@params` that document both exceptions thrown, namely `IllegalArgumentException` if the signal is already registered, and `UnsupportedOperationException` if the OS isn't supported.

server/src/main/scala/de/frosner/broccoli/signal/UnixSignalManager.scala
```diff
@@ -0,0 +1,32 @@
+package de.frosner.broccoli.signal
+
+import org.apache.commons.lang3.SystemUtils
+import sun.misc.{Signal, SignalHandler}
+
+import scala.collection.mutable
+
+class UnixSignalManager extends SignalManager {
+  private val signals = mutable.HashMap.empty[Signal, SignalHandler]
```
swsnr: Does `Signal` hash by signal name?   IOW, does `new Signal("USR2") == new Signal("USR2")` hold?

server/src/test/scala/de/frosner/broccoli/signal/UnixSignalManagerSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+      manager.register(new Signal("USR1"), mock[SignalHandler]) must throwA(
+        new IllegalArgumentException("Signal already used by VM or OS: SIGUSR1"))
+    }
+
+    "fail if the signal is already registered" in {
+      val manager = new UnixSignalManager()
+      val signal = new Signal("USR2")
+      val handler = mock[SignalHandler]
+      manager.register(signal, handler)
+      manager.register(signal, handler) must throwA(
```
swsnr: Please rewrite the test to explicitly create a new with `new Signal("USR2")` in both `.register` calls, to make sure that two distinct Signal objects for the same signal trigger the exception.

server/src/main/scala/de/frosner/broccoli/signal/UnixSignalManager.scala
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+  def unregister(signal: Signal): Unit =
+    if (signals.contains(signal)) {
+      Signal.handle(signal, new SignalHandler {
+        override def handle(signal: Signal): Unit = {}
+      })
+      signals.remove(signal)
+    } else {
+      throw new IllegalArgumentException(s"Signal $signal is not registered")
```
swsnr: I'd not throw an exception here; unregistering a signal that is not registered is not a failure imho but rather an idempotent operation.

swsnr (changes requested):

server/src/test/scala/de/frosner/broccoli/templates/SignalRefreshedTemplateSourceSpec.scala
```diff
@@ -8,8 +9,9 @@ import sun.misc.Signal
 class SignalRefreshedTemplateSourceSpec extends Specification with Mockito {
   "Receiving a SIGUSR2 signal" should {
     "update the cache" in {
+      val signalManager = new UnixSignalManager()
```
swsnr: Now that the signal manager moved out of the cached template source, please rewrite this test with a mocked signal manager so that it doesn't need the actual USR2 signal anymore.

server/src/test/scala/de/frosner/broccoli/signal/UnixSignalManagerSpec.scala
```diff
@@ -0,0 +1,33 @@
...
+class UnixSignalManagerSpec extends Specification with Mockito {
+  "Registering new signal" should {
+    "trigger the handler when the signal is raised" in {
+      val manager = new UnixSignalManager()
+      val signal = new Signal("USR2")
+      val handler = mock[SignalHandler]
+      manager.register(signal, handler)
+      Signal.raise(signal)
+      Thread.sleep(1000)
+      there was one(handler).handle(signal)
```
swsnr: Please turn this into an integration test.

server/src/main/scala/de/frosner/broccoli/signal/SignalModule.scala
```diff
@@ -0,0 +1,14 @@
+package de.frosner.broccoli.signal
+
+import javax.inject.Singleton
+
+import com.google.inject.{AbstractModule, Provides}
+import net.codingwell.scalaguice.ScalaModule
+
+class SignalModule extends AbstractModule with ScalaModule {
```
swsnr: I think you forgot to register this module in `application.conf` didn't you?

server/src/main/scala/de/frosner/broccoli/signal/SignalModule.scala
```diff
@@ -0,0 +1,14 @@
...
+  override def configure(): Unit = {}
+
+  @Provides
+  @Singleton
+  def provideSignalManager(): UnixSignalManager = new UnixSignalManager()
```
swsnr: Please use a simple `.bind` expression in `configure`; we don't need a provider method here.

server/src/main/scala/de/frosner/broccoli/signal/SignalManager.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import sun.misc.{Signal, SignalHandler}
+
+/**
+  * Provide a way to register and unregister OS signals
+  */
+trait SignalManager {
+
+  /**
+    * Registers signal handler for the given signal
```
swsnr: Minor nitpick: Please add a punctuation at the end, and a blank line below the first line.

In Scaladoc the first line serves as a summary that you see in the overview list of all methods. This only works if there's a blank line after it, and it reads better with proper punctuation.

server/src/main/scala/de/frosner/broccoli/signal/SignalManager.scala
```diff
@@ -0,0 +1,24 @@
...
+    * @param signal
+    * @param handler
+    * @throws IllegalArgumentException if a signal is already registered or reserved by JDK or OS
+    * @throws UnsupportedOperationException if OS is not supported
+    */
+  def register(signal: Signal, handler: SignalHandler): Unit
+
+  /**
+    * Unregisters the signal
```
swsnr, line 21: Likewise here.

swsnr (changes requested):

swsnr (approved):

swsnr:
@Gerrrr Will merge as soon as the build passed :)

Gerrrr:
@lunaryorn cheers :)

swsnr:
Incredible

Issue #17, opened 2016-06-24
Input sanitization


FRosner:
### Problem

When the template JSON is built in `Instance`, people might be able to "inject" JSON. We might want to allow to configure an input sanitization (e.g. escaping quotes or other JSON characters). However as a user might want to allow JSON inside we should make it configurable whether you want to sanitize and how (e.g. escape quotes).

### Solution

Make sanitization configurable. Should it be on a global level (all input), on a template-parameter level or both? => I think best is to actually treat text parameters as JSON text, which sanitizes it automatically. Thus when you can put `"key" : {{value}}` instead of `"key" : "{{value}}"` if `{{value}}` is a text parameter.

And then we can just add another parameter type which we call "raw" in addition to "string". This relates to #71 also.

- [ ] Introduce a new config object for templates and instances (like it is done for nomad in https://github.com/FRosner/cluster-broccoli/pull/254/files#diff-44c416010933ac7e004b47d735296d38)
- [ ] Important: If we make "string" the default parameter type then it will break all existing templates. However I think that "string" is the most reasonable default because it is safe. To avoid putting "raw" as default we can either 1) encode the Broccoli version into the template somehow and when we detect that there is no version present we choose "raw" as default, otherwise "string", or 2) add a global configuration parameter to change the default so we can allow users to have some deprecation period.
- [ ] Refactor instance parsing so I can inject the config option which parameter type to use as default.

FRosner:
@Gerrrr this is my current status. As discussed it makes sense to factor this logic out of instance.

```
diff --git a/server/src/main/scala/de/frosner/broccoli/models/Instance.scala b/server/src/main/scala/de/frosner/broccoli/models/Instance.scala
index 47fb0b4..ff32838 100644
--- a/server/src/main/scala/de/frosner/broccoli/models/Instance.scala
+++ b/server/src/main/scala/de/frosner/broccoli/models/Instance.scala
@@ -47,16 +47,25 @@ case class Instance(id: String, template: Template, parameterValues: Map[String,
     }

   def templateJson: JsValue = {
+    val parameterInfos = template.parameterInfos
+    def sanitize(parameter: String, value: String) = {
+      val parameterType = parameterInfos.get(parameter).flatMap(_.`type`).getOrElse(ParameterType.Raw) // TODO @lunaryorn get this from config
+      val sanitized = parameterType match {
+        case ParameterType.Raw    => value
[...]

PR #265, opened 2017-08-03
Input sanitization

Gerrrr:
Solves #17


server/src/main/scala/de/frosner/broccoli/instances/InstanceModule.scala
```diff
@@ -0,0 +1,70 @@
...
+      case StorageType.FileSystem => {
+        val config = instanceConfig.storageConfiguration.fsConfig
+        val path = FileSystems.getDefault.getPath(config.url).toAbsolutePath
+        Try(FileSystemInstanceStorage(path.toFile))
+      }
+      case StorageType.CouchDB => {
+        val config = instanceConfig.storageConfiguration.couchDBConfig
+        Try(CouchDBInstanceStorage(config.url, config.dbName, ws))
+      }
+      case invalidStorageType => throw new IllegalArgumentException("Wrong instance storage type")
```
swsnr: Isn't this case unreachable, given that `StorageType` has just two fixed variants?

server/src/main/scala/de/frosner/broccoli/instances/InstanceModule.scala
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    val instanceStorage = maybeInstanceStorage match {
+      case Success(storage) => storage
+      case Failure(throwable) =>
+        Logger.error(s"Cannot start instance storage: $throwable")
+        throw throwable
+    }
```
swsnr: I'm not sure that we really need to use `Try` here given that we just re-throw the exception.

I'd just remove `Try`, create the instance storages directly and let exceptions propagate.  They'll end up in the logger anyway.

server/src/main/scala/de/frosner/broccoli/instances/InstanceModule.scala
```diff
@@ -0,0 +1,70 @@
...
+import play.api.inject.ApplicationLifecycle
+import play.api.libs.ws.WSClient
+
+import scala.concurrent.Future
+import scala.util.{Failure, Success, Try}
+
+/**
+  * Provide Instance and InstanceStorage configuration from the Play configuration
+  */
+class InstanceModule extends AbstractModule with ScalaModule with Logging {
```
swsnr: Do not use the `Logging` trait; I think it's redundant, and I'd like to deprecate it.

Instead just add `private val log = Logger(getClass)` which will create a nicely configured logger automatically.  Then replace all instances of `Logger` in this module with `log`.

server/src/main/scala/de/frosner/broccoli/instances/StorageType.scala
```diff
@@ -0,0 +1,16 @@
+package de.frosner.broccoli.instances
+
+import enumeratum.{Enum, EnumEntry, PlayJsonEnum}
+
+/**
+  * Type of the instance storage
+  *
+  * @param entryName type name used in the Play configuration
+  */
+sealed abstract class StorageType(override val entryName: String) extends EnumEntry with EnumEntry.Lowercase
```
swsnr: Since you're explicitly setting the entry names, ie, have no auto-generated names, you don't need to use `EnumEntry.Lowercase` here.  Just remove it.

server/src/main/scala/de/frosner/broccoli/instances/StorageType.scala
```diff
@@ -0,0 +1,16 @@
...
+
+object StorageType extends Enum[StorageType] with PlayJsonEnum[StorageType] {
+  val values = findValues
```
swsnr: Please add a type annotation (ie, press `alt + enter`, then select "add type annotation") and `override` here.

server/src/main/scala/de/frosner/broccoli/instances/conf/CouchDBInstanceStorageConfiguration.scala
```diff
@@ -0,0 +1,23 @@
...
+  *
+  * @param url database address
+  * @param dbName database name where the instance information will be stored
+  */
+final case class CouchDBInstanceStorageConfiguration(url: String, dbName: String)
+
+object CouchDBInstanceStorageConfiguration extends Logging {
+  def fromConfig(config: Config): CouchDBInstanceStorageConfiguration = {
+    val url = config.getString("url")
+    Logger.info(s"broccoli.instances.storage.couchdb.url=$url")
```
swsnr: A simple factory method like this shouldn't have side-effects, not even logging.  Please move these logging calls to the couchdb storage class or to the instances module (wherever they fit best).

server/src/main/scala/de/frosner/broccoli/instances/conf/FileSystemInstanceStorageConfiguration.scala
```diff
@@ -0,0 +1,19 @@
...
+  * Configuration for FileSystemInstanceStorage
+  *
+  * @param url location on the filesystem to store the instance information
+  */
+final case class FileSystemInstanceStorageConfiguration(url: String)
+
+object FileSystemInstanceStorageConfiguration extends Logging {
+  def fromConfig(config: Config): FileSystemInstanceStorageConfiguration = {
+    val url = config.getString("url")
+    Logger.info(s"broccoli.instances.storage.fs.url=$url")
```
swsnr: Likewise, please move this logging call to the appropriate place (see above).

server/src/main/scala/de/frosner/broccoli/instances/conf/FileSystemInstanceStorageConfiguration.scala
```diff
@@ -0,0 +1,19 @@
...
+import de.frosner.broccoli.util.Logging
+
+/**
+  * Configuration for FileSystemInstanceStorage
+  *
+  * @param url location on the filesystem to store the instance information
+  */
+final case class FileSystemInstanceStorageConfiguration(url: String)
+
+object FileSystemInstanceStorageConfiguration extends Logging {
```
swsnr: Please remove the logging trait here.

server/src/main/scala/de/frosner/broccoli/instances/conf/CouchDBInstanceStorageConfiguration.scala
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+  * Configuration for CouchDBInstanceStorage
+  *
+  * @param url database address
+  * @param dbName database name where the instance information will be stored
+  */
+final case class CouchDBInstanceStorageConfiguration(url: String, dbName: String)
+
+object CouchDBInstanceStorageConfiguration extends Logging {
```
swsnr: Please remove the logging trait here.

server/src/main/scala/de/frosner/broccoli/instances/conf/InstanceConfiguration.scala
```diff
@@ -0,0 +1,33 @@
...
+  *
+  * @param pollingFrequency Integer (seconds) to control the time between asking Nomad and Consul for job and service status
+  * @param defaultParameterType The parameter type to be used by default for the instance parameters
+  * @param storageConfiguration Configuration specific to the instance storage type
+  */
+final case class InstanceConfiguration(pollingFrequency: Long,
+                                       defaultParameterType: ParameterType,
+                                       storageConfiguration: InstanceStorageConfiguration)
+
+object InstanceConfiguration extends Logging {
```
swsnr: Please remove the logging trait here.

server/src/main/scala/de/frosner/broccoli/instances/conf/InstanceConfiguration.scala
```diff
@@ -0,0 +1,33 @@
...
+  def fromConfig(config: Config): InstanceConfiguration = {
+    val pollingFrequency = config.getLong("polling.frequency")
+    if (pollingFrequency <= 0) {
+      throw new IllegalArgumentException(
+        s"Invalid polling frequency specified: $pollingFrequency. Needs to be a positive integer.")
+    }
+
+    Logger.info(s"Nomad/Consul polling frequency set to $pollingFrequency seconds")
```
swsnr: A simple factory method like this shouldn't have side-effects, not even logging. Please move these logging calls to the couchdb storage class or to the instances module (wherever they fit best).

server/src/main/scala/de/frosner/broccoli/instances/conf/InstanceStorageConfiguration.scala
```diff
@@ -0,0 +1,31 @@
...
+  *
+  * @param storageType The storage type to use. Can be either 'fs' or 'couchdb'
+  * @param fsConfig  Configuration for FileSystemInstanceStorage
+  * @param couchDBConfig Configuration for CouchDBInstanceStorage
+  */
+final case class InstanceStorageConfiguration(storageType: StorageType,
+                                              fsConfig: FileSystemInstanceStorageConfiguration,
+                                              couchDBConfig: CouchDBInstanceStorageConfiguration)
+
+object InstanceStorageConfiguration extends Logging {
```
swsnr: Please remove the `Logging` trait here, you're not using it in this object.

server/src/main/scala/de/frosner/broccoli/models/Instance.scala
```diff
@@ -10,8 +9,8 @@ case class Instance(id: String, template: Template, parameterValues: Map[String,
 
   def requireParameterValueConsistency(parameterValues: Map[String, String], template: Template) = {
     val realParametersWithValues = parameterValues.keySet ++ template.parameterInfos.flatMap {
-      case (key, ParameterInfo(_, _, Some(default), _)) => Some(key)
-      case (key, ParameterInfo(_, _, None, _))          => None
+      case (key, ParameterInfo(_, _, Some(default), _, _)) => Some(key)
+      case (key, ParameterInfo(_, _, None, _, _))          => None
```
swsnr: Replace with `case (key, info) => info.default.map(Function.const(key))`; that's more straight-forward and easier to read than the complex destructuring in the pattern match.

server/src/main/scala/de/frosner/broccoli/models/ParameterType.scala
```diff
@@ -0,0 +1,11 @@
+package de.frosner.broccoli.models
+
+import enumeratum.{Enum, EnumEntry, PlayJsonEnum}
+
+sealed trait ParameterType extends EnumEntry with EnumEntry.Lowercase
+
+object ParameterType extends Enum[ParameterType] with PlayJsonEnum[ParameterType] {
```
swsnr: Please add API documentation to this type.

Furthermore please remove `PlayJsonEnum`; we never use this enum in JSON, do we?

server/src/main/scala/de/frosner/broccoli/instances/StorageType.scala
```diff
@@ -0,0 +1,16 @@
...
+import enumeratum.{Enum, EnumEntry, PlayJsonEnum}
+
+/**
+  * Type of the instance storage
+  *
+  * @param entryName type name used in the Play configuration
+  */
+sealed abstract class StorageType(override val entryName: String) extends EnumEntry with EnumEntry.Lowercase
+
+object StorageType extends Enum[StorageType] with PlayJsonEnum[StorageType] {
```
swsnr: Please remove `PlayJsonEnum` here; you never use this enum in JSON, do you?

server/src/main/scala/de/frosner/broccoli/templates/TemplateRenderer.scala
```diff
@@ -0,0 +1,45 @@
...
+  *
+  * @param instanceConfiguration
+  */
+class TemplateRenderer @Inject()(instanceConfiguration: InstanceConfiguration) {
+  def renderJson(instance: Instance): JsValue = {
+    val template = instance.template
+    val parameterInfos = template.parameterInfos
+    val parameterValues = instance.parameterValues
+
+    def sanitize(parameter: String, value: String) = {
```
swsnr: Please move this to a top-level function in `TemplateRenderer` so that you can test it independently.

server/src/main/scala/de/frosner/broccoli/instances/InstanceModule.scala
```diff
@@ -0,0 +1,70 @@
...
+    }
+    implicit val executionContext = play.api.libs.concurrent.Execution.Implicits.defaultContext
+    applicationLifecycle.addStopHook(() =>
+      Future {
+        Logger.info("Closing instanceStorage (stop hook)")
+        if (!instanceStorage.isClosed) {
+          instanceStorage.close()
+        }
+    })
+    instanceStorage
```
swsnr: I wonder why this code uses a `sys.addShutdownHook` in addition to a normal stop hook.  Can you try to find out?

I think we should remove the shutdown hook and just leave the stop hook; it should be sufficient.

I'm also curious why closing the storage runs on a separate thread via `Future.apply` here.  I suggest to just close it right away without wrapping it in a `Future` and then return `Future.successful(())` at the end. The documentation of `addStopHook` explicitly allows this, and we gain nothing with a separate thread here.

swsnr (changes requested):

.gitignore
```diff
@@ -16,3 +16,6 @@ target/
 
 # results from HTTP API testing tool
 *run-*
+
+# Intelij Idea files
+.idea/*
```
swsnr: A `.gitignore` in a repo should normally only contain files being created by the repository itself, ie, build artifacts or local caches.

For IDE metadata the convention is to keep them in your [global ignore file](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).  

I'd appreciate if you could remove `.idea/` from `.gitignore` again and move it to your own global ignore file.

swsnr (changes requested):

.gitignore
Gerrrr: Hey @lunaryorn , 
Thanks for the suggestion, I was not aware of the global ignore file!

swsnr:
@Gerrrr Tell me when you're ready and I'll review.  Perhaps tomorrow?

server/src/main/scala/de/frosner/broccoli/instances/conf/InstanceConfiguration.scala
```diff
@@ -0,0 +1,35 @@
...
+  private val log = Logger(getClass)
+
+  def fromConfig(config: Config): InstanceConfiguration = {
+    val pollingFrequency = config.getLong("polling.frequency")
+    if (pollingFrequency <= 0) {
+      throw new IllegalArgumentException(
+        s"Invalid polling frequency specified: $pollingFrequency. Needs to be a positive integer.")
+    }
+
+    log.info(s"Nomad/Consul polling frequency set to $pollingFrequency seconds")
```
swsnr: Please remove this logging statement; it doesn't belong here.

If we need to log this for some reason, the class that's using this setting should log it.

server/src/main/scala/de/frosner/broccoli/services/InstanceService.scala
```diff
@@ -25,26 +24,12 @@ class InstanceService @Inject()(nomadClient: NomadClient,
...
-    case None => Success(conf.POLLING_FREQUENCY_DEFAULT)
-  }
-  if (pollingFrequencySecondsTry.isFailure) {
-    Logger.error(
-      s"Invalid ${conf.POLLING_FREQUENCY_KEY} specified: '${pollingFrequencySecondsString.get}'. Needs to be a positive integer.")
-    System.exit(1)
-  }
-  private lazy val pollingFrequencySeconds = pollingFrequencySecondsTry.get
-  Logger.info(s"Nomad/Consul polling frequency set to $pollingFrequencySeconds seconds")
+  private lazy val pollingFrequencySeconds = instanceConfiguration.pollingFrequency
```
swsnr: The instance service uses the polling frequency, so if at all, the instance service should log it.

server/src/main/scala/de/frosner/broccoli/templates/TemplateRenderer.scala
```diff
@@ -0,0 +1,45 @@
...
+import de.frosner.broccoli.models.{Instance, ParameterInfo, ParameterType}
+import play.api.libs.json.{JsString, JsValue, Json}
+
+/**
+  * Renders json representation of the passed instance
+  *
+  * @param instanceConfiguration
+  */
+class TemplateRenderer @Inject()(instanceConfiguration: InstanceConfiguration) {
+  private[templates] def sanitize(parameter: String, value: String, parameterInfos: Map[String, ParameterInfo]) = {
```
swsnr: Why `private[templates]`?

swsnr (changes requested):
A few minor changes, but other than that looks good :+1:

server/src/main/scala/de/frosner/broccoli/templates/TemplateRenderer.scala
Gerrrr: Because it is a function that it is only used by `renderJson` and not supposed to be public, but I still want to unit test it.

FRosner:
Hi @Gerrrr. I'd appreciate if we changed the templates in a way that there is at least one example of a real string parameter. Right now all of them are raw but maybe we can add some just for the sake of an example?

server/src/main/scala/de/frosner/broccoli/templates/TemplateRenderer.scala
swsnr: Why not just make it public?

swsnr:
@Gerrrr I think you need to rebase again; I merged #278 which seems to have triggered a merge conflicts.

Please ping me when you're done and I'll approve this PR.

swsnr (dismissed):

server/src/main/resources/reference.conf
```diff
@@ -1,4 +1,16 @@
...
+    storage {
+      type = "fs"
+      fs.url = "./instances"
+      couchdb {
+        url = "http://localhost:5984"
+        dbName = "broccoli_instances"
+      }
+    }
+    parameters.defaultType = "string"
+  }
```
FRosner, line 15: - [x] Now that we have these changes in the `reference.conf` I think it is time to either add the new parameter to the README or change the README to point to the `reference.conf` for configuration reference and remove the long tables I built. If we do that, we should at least put some comments though, so we don't lose any information that is currently provided.
- [ ] We should also add a section to the Wiki migration guide to 0.7.0, explaining the defaultType. 
- [ ] The new option in `meta.json` should also be explained in the Wiki, no?

FRosner (changes requested):

server/src/main/scala/de/frosner/broccoli/services/InstanceService.scala
```diff
@@ -26,17 +27,26 @@ class InstanceService @Inject()(nomadClient: NomadClient,
...
-  private lazy val pollingFrequencySeconds = instanceConfiguration.pollingFrequency
+  // FIXME: refactor out together with the polling scheduler
+  private lazy val pollingFrequencySeconds = {
+    val pollingFrequency = config.getLong("polling.frequency").get
+    if (pollingFrequency <= 0) {
+      throw new IllegalArgumentException(
+        s"Invalid polling frequency specified: $pollingFrequency. Needs to be a positive integer.")
+    }
+    pollingFrequency
+  }
```
swsnr: Uh, why did you move this back?!  You already fixed it, and now unfixed it?

swsnr (changes requested):

server/src/main/scala/de/frosner/broccoli/services/InstanceService.scala
swsnr: Ah, I see… you removed it from the instance configuration.  Fine with me then.

swsnr (dismissed):

server/src/main/resources/reference.conf
Gerrrr: I've added the new parameter to the README. I will update wiki after merging the PR and before closing the issue. Does it make sense?

lunaryorn:
@Gerrrr Sure, as far as I'm concerned.

server/src/main/resources/reference.conf
Gerrrr: yes

swsnr (approved):

FRosner (approved):

PR #266, opened 2017-08-07
Strengthen type of webservice messages

swsnr:
On the Scala side replace any-typed payload with sealed trait enum, and move
from iterable to Seq.

On the Elm side define websocket messages as a proper ADT instead of attaching
types to JSON blobs.  Decode messages as early and encode as late as possible,
and use typesafe message representation internally.

Greatly improves type safety, reduces error cases, and overall simplifies the
code.

Chore work for #44


FRosner (approved):
LGTM

PR #269, opened 2017-08-08
Add HTTP client to get allocations from Nomad API

swsnr:
Runs integration tests for the Nomad client against the docker test image from specs2, i.e., adds infrastructure to gradually replace the bash-scripted integration tests.


FRosner (approved):
LGTM

PR #271, opened 2017-08-08
Make instance errors generic

swsnr:
Replace explicit result enums per operation type, e.g., `InstanceUpdateResult` with an `Either` over a newly introduced `InstanceError` and the old "success" type, and rename the success types to remove the now redundant "success" from their name.

Define the new `InstanceError` as ADT with specialised variants for all previously known error cases of instance operations.

Rewrite instance operations to make use of the monadic nature of `Either`, thus reducing nesting and branching to make the code more straightforward and simpler.  

In a related change explicitly specify the body parsers of all controller methods to shift the work of parsing and validating JSON to Play, and thus removing boilerplate from the controller actions, which makes all controller actions trivial wrappers around the controller operations.

In an unrelated change turn `Role` into an `enumerated` enum just because I can 😎


server/src/main/scala/de/frosner/broccoli/models/InstanceError.scala
```diff
@@ -0,0 +1,101 @@
...
+  object RolesRequired {
+    def apply(role: Role, roles: Role*): RolesRequired = RolesRequired(Set(role) ++ roles)
+  }
+
+  /**
+    * A generic error thrown while operating on an instance.
+    *
+    * @param throwable The throwable thrown by the operation
+    */
+  @deprecated(message = "Do not use, map throwables to specific instance errors or propagate them to trigger ISEs",
```
FRosner, line 82: What are ISEs?

server/src/main/scala/de/frosner/broccoli/models/InstanceError.scala
```diff
@@ -0,0 +1,101 @@
...
+    * @param reason A description of why parameters were invalid
+    */
+  final case class InvalidParameters(reason: String) extends InstanceError
+
+  /**
+    * A user was denied to perform an operation on an instance because their instance regex didn't allow access to the
+    * instance.
+    *
+    * @param instanceId The ID of the instance
+    * @param user The user who tried to access the instance
```
FRosner: ScalaDoc does not match the parameters

server/src/main/scala/de/frosner/broccoli/controllers/InstanceController.scala
```diff
@@ -51,54 +52,28 @@ case class InstanceController @Inject()(instanceService: InstanceService, overri
...
-            }
-          }
-          .getOrElse(Results.Status(400)("Expected JSON data"))
-      }
-      .getOrElse(Results.Status(400)("Expected JSON data"))
+  def create = StackAction(parse.json[InstanceCreation]) { implicit request =>
+    InstanceController
+      .create(request.body, loggedIn, instanceService)
+      .fold(
+        _.toHTTPResult,
```
FRosner: How does Scala know how to convert the error into an HTTP result? I see the ToHTTPResult typeclass but I'm not sure I understand how it works.

FRosner:
You changed `user` to `normaluser`. Thus using the default `application.conf` breaks:

```
17:03:48.246 ERROR SecurityService - Error parsing broccoli.auth.conf.accounts: java.util.NoSuchElementException: user is not a member of Enum (administrator, operator, normaluser)
```

If we make this change we need to document it in the migration guide, adjust the documentation in the wiki and adjust the config as well. To track this I prefer creating a ticket because I am going through the tickets to see if any change in the Wiki is required when releasing.

Is there a strong reason to rename it?

server/src/main/scala/de/frosner/broccoli/models/InstanceError.scala
swsnr: "Internal Server Error", ie, HTTP 500 responses.

swsnr:
It was `NormalUser` before; I just didn't realise that the entry names were different from the enum member names.  Arguably that wasn't the best idea to begin with.  I'll rename to `User`.

Edit: Renamed to `User`.

server/src/main/scala/de/frosner/broccoli/controllers/InstanceController.scala
swsnr: The `@typeclass` annotation generates boilerplate for type classes (see [README](https://github.com/mpilquist/simulacrum)), including an `ops` object which provides syntax support (hence the "ops" import above).  Essentially the "ops" object turns every method of the type class into an extension method on values that have an instance of the type class available.

And indeed, `InstanceError` has a `ToHTTPResult` instance in its companion object.

server/src/main/scala/de/frosner/broccoli/models/InstanceError.scala
swsnr: Fixed.

FRosner (commented):
Please check why the start stop buttons disappear when you are admin.

swsnr:
@FRosner I fixed the role issue; there was a "role.toString" at some places, which doesn't work with enumeratum.  The code's using the real Role type now, and doesn't pass roles around as strings.

The UI works now, please review again :)

FRosner:
@lunaryorn where did you add this? The last commit still seems to be the same?

FRosner (approved):

swsnr:
@FRosner Thanks <3

swsnr:
@FRosner I amended the "Turn Role in nice enum" commit.

Issue #272, opened 2017-08-09
Extract secrets filtering into type class


swsnr:

swsnr:
Closes by #274

PR #273, opened 2017-08-10
Send tasks over web sockets

swsnr:
* Add websocket messages for instance tasks to server and web
* Turn processing of incoming messages into an Enumeratee flow over Future.  The server retrieves instance tasks asynchronously, so use an Enumeratee mapping over a `Future` to define the Iteratee for incoming messages.
* Move dispatching of messages, ie, mapping incoming to outgoing messages, into a separate component to tear apart different responsibilities and make processing testable without elaborate websocket setup.

PR #274, opened 2017-08-10
Extract instance filtering into a type class

swsnr:
See GH-272


FRosner:
I'm getting

```
[warn] /Users/frosneraz/Documents/projects/cluster-broccoli/server/src/main/scala/de/frosner/broccoli/RemoveSecrets.scala:10: implicit conversion method toRemoveSecretsOps should be enabled
[warn] by making the implicit value scala.language.implicitConversions visible.
[warn] This can be achieved by adding the import clause 'import scala.language.implicitConversions'
[warn] or by setting the compiler option -language:implicitConversions.
[warn] See the Scaladoc for value scala.language.implicitConversions for a discussion
[warn] why the feature should be explicitly enabled.
[warn] @typeclass trait RemoveSecrets[T] {
[warn]  ^
```

Should we do what the compiler says? What does it mean?

swsnr:
@FRosner Yes, we should do as the compiler says; I just forgot.

A `scala.language` feature import enables specific language features that have complex side-effects, eg, turing-completeness for higher kinded types, or have a certain potential for abuse, like implicit conversions.

The `scala.language` [scaladoc](http://www.scala-lang.org/api/2.11.8/#scala.language$) has a list of all flags, their meanings and the reasons why they exist.  Wrt `implicitConversions`:

> **Why keep the feature?** Implicit conversions are central to many aspects of Scala’s core libraries.
>
> **Why control it?** Implicit conversions are known to cause many pitfalls if over-used. And there is a tendency to over-use them because they look very powerful and their effects seem to be easy to understand. Also, in most situations using implicit parameters leads to a better design than implicit conversions.

The macro behind the `@typeclass` annotation triggers this warning because it generates an implicit conversion to the `ops` trait (remember what we discussed on the `ToHTTPResults` PR?)

swsnr:
@FRosner Added missing warning and rebase on top of #273.

FRosner (approved):

PR #275, opened 2017-08-10
Show tasks and instances in the web UI

swsnr:


FRosner:
"Tasks and allocations of this instance" should be formatted like the other captions (parameters and template). Also "of this instance" is redundant information as we already know what we are looking at when expanding the instance. So I think "Allocations" is maybe the most concise way to describe what we are looking at?

FRosner:
If there are no allocations it should either not show this section at all or state below the "Allocations" caption that there are no allocations available.

Currently the "Periodic runs" section is just not visible so we should anyway be consistent. Either show both or hide both if there is nothing to show, IMHO.

FRosner (changes requested):
I think we need to address the small issues mentioned in the comments but other than that LGTM

FRosner:
Also do you think it makes sense to show only the prefix of the allocation ID? I.e. `00910091` instead of `00910091-0dd7-f8c0-adb0-54f6575402c2`? It's what is commonly used to interact with the Nomad command line and for people without command line access the number is not so important.

swsnr:
@FRosner I'll use the short ID, but I'd like to discuss the labels.

> So I think "Allocations" is maybe the most concise way to describe what we are looking at?

Will someone who's not familiar with Nomad understand what an "allocation" is?  I don't think so… "tasks" at least has a generic meaning that more or less fits what we display in this section.

I'm fine with omitting "this instance".

> Currently the "Periodic runs" section is just not visible so we should anyway be consistent.

Consistency is no inherent goal.  I think an instance not having any tasks is helpful information; stating it explicitly makes it easier to see and to understand.

FRosner:
- @lunaryorn the short ID is nicer and we can later add that clicking on it or on a button next to it will copy the full id.

- About the allocation you are right. How about 'Task Allocations'? I want to make sure that it is clear that these are actual allocated tasks and not just tasks. Because when they are not allocated they are not shown and I find this confusing.

- I agree with showing an empty allocations field. Should we then also show that there are no period runs? I just want to avoid that something pops up that you might not be expecting. What do you think?

swsnr:
@FRosner Does the UI explain or indicate the different between "allocated" and "unallocated" tasks?  Can we explain this difference in a simple way?  Do we want the user to care for this difference?  All in all, does "as a Broccoli user I'd like to see the difference between allocated and unallocated tasks" sound like anything we'd ever hear?

I do not know where periodic runs come from; depending on that it might make sense to show this information.  I'd show this information if it's not obvious that an instance has no periodic runs.

swsnr:
Change the labels to "allocated tasks", use h5 for headline, and shorten the allocation ID:

new ui

Anything else?

FRosner:
I have troubles getting it to run. Sometimes the UI does not respond to my actions and sometimes the `yarn start` version of the UI is inconsistent with the normal one. Do you have a minute to look at it?

FRosner (commented):
LGTM. Can you check my latest change if you are ok with it? Should we test this functionality?

Thanks!

webui/src/Views/InstanceView.elm
```diff
@@ -363,30 +363,37 @@ getActiveAllocations task =
...
+                                ]
                             ]
+                        , tbody [] <| List.indexedMap instanceAllocationRow allocations
                         ]
-                    , tbody [] <| List.indexedMap instanceAllocationRow allocations
                     ]
+    in
+        div
+            [ style
+                [ ( "margin-bottom", "15px" ) ]
```
swsnr: Why do we need this?

FRosner: Otherwise it looks ugly, because the content of the allocation div (e.g. "No tasks have been allocated, yet.") is stuck to the next elements on the bottom.

image

swsnr: The screenshot doesn't look ugly to me—is it with or without margin?  

In any case please don't use ad-hoc in-line styles.  We should either use proper classes, or at  define styles in Elm records at a central place.

swsnr: Furthermore please add the margin to the top-level container of the expanded instance, not to an arbitrary div in between, so that we can safely add new elements to the expander without breaking the layout.

FRosner:
The screenshot has the margin added already, yes. I agree with your point about using classes or centrally defined styles.

swsnr:
@FRosner Will you make the corresponding changes to this PR?

FRosner:
Done. Can you check?

swsnr:
@FRosner :+1:

Anything still missing?

FRosner (approved):

PR #276, opened 2017-08-14
GH-44: Add HTTP API endpoint to get instance tasks

swsnr:


FRosner:
We need to document this endpoint in the Wiki. Can you provide some documentation similar to the one available already: https://github.com/FRosner/cluster-broccoli/wiki/HTTP-API-v1?

swsnr:
@FRosner Shouldn't I do that *after* merging?  Why do we not move the documentation into the repository itself (which is where it belongs in my opinion)?

FRosner:
Agreed we should move it. I wanted to have it here before merging so I can review it. But if it is part of the code that would be even better. Do you need any support?

swsnr:
I'll add docs.

swsnr:
@FRosner Added docs: https://github.com/FRosner/cluster-broccoli/wiki/HTTP-API-v1#apiv1instancesinstanceidtasks

FRosner:
The curl command in the docs seems wrong? `curl -X GET 'localhost:9000/api/v1/instances/`. I'm going to update it.

swsnr:
@FRosner Wrong? In how far?  Please be specific.

FRosner:
First of all it is missing a quote.

image

After adding it, I'm getting an error that the route cannot be found. Which is because it is wrong, no? It should be `curl -X GET 'localhost:9000/api/v1/instances/my-http/tasks'`

FRosner (commented):
LGTM

FRosner (approved):

swsnr:
@FRosner Ah, my bad.  Sorry, and thanks for fixing.

PR #277, opened 2017-08-16
Fix compiler warnings

swsnr:
Clean compile before:

```
[info] Compiling 84 Scala sources and 2 Java sources to /Users/swiesner/GitHub/FRosner/cluster-broccoli/server/target/scala-2.11/classes...
[warn] /Users/swiesner/GitHub/FRosner/cluster-broccoli/server/src/main/scala/de/frosner/broccoli/services/InstanceService.scala:54: implicit numeric widening
[warn]   private val scheduledTask = scheduler.scheduleAtFixedRate(task, 0, pollingFrequencySeconds, TimeUnit.SECONDS)
[warn]                                                                      ^
[warn] /Users/swiesner/GitHub/FRosner/cluster-broccoli/server/src/main/scala/Global.scala:6: trait GlobalSettings in package api is deprecated: Use dependency injection
[warn] object Global extends GlobalSettings {
[warn]                       ^
[warn] /Users/swiesner/GitHub/FRosner/cluster-broccoli/server/src/main/scala/de/frosner/broccoli/controllers/AuthConfigImpl.scala:64: method maybeApplication in object Play is deprecated: This is a static reference to application, use DI instead
[warn]     cookieSecureOption = play.api.Play.maybeApplication.exists(app => play.api.Play.isProd(app) && cookieSecure),
[warn]                                        ^
[warn] /Users/swiesner/GitHub/FRosner/cluster-broccoli/server/src/main/scala/de/frosner/broccoli/controllers/AuthConfigImpl.scala:64: method isProd in object Play is deprecated: inject the play.api.Environment instead
[warn]     cookieSecureOption = play.api.Play.maybeApplication.exists(app => play.api.Play.isProd(app) && cookieSecure),
[warn]                                                                                     ^
[warn] /Users/swiesner/GitHub/FRosner/cluster-broccoli/server/src/main/scala/de/frosner/broccoli/controllers/WebSocketController.scala:66: method tryAccept in object WebSocket is deprecated: Use acceptOrResult with an Akka streams flow instead
[warn]   def socket: WebSocket = WebSocket.tryAccept[Msg](requestToSocket)
[warn]                                     ^
[warn] /Users/swiesner/GitHub/FRosner/cluster-broccoli/server/src/main/scala/de/frosner/broccoli/services/FileSystemInstanceStorage.scala:51: non-nullary method overrides nullary method
[warn]   protected override def readInstancesImpl(): Try[Set[Instance]] =
[warn]                          ^
[...]

Issue #81, opened 2016-09-20
Better error message for invalid options


omame:
When a user starts the app with an invalid option (in my case I forgot to add `http://` before the nomad url) this is what gets reported:

```
[error] a.a.OneForOneStrategy - scheme
java.lang.NullPointerException: scheme
        at com.ning.http.client.uri.Uri.<init>(Uri.java:56) ~[com.ning.async-http-client-1.9.21.jar:na]
        at com.ning.http.client.uri.Uri.create(Uri.java:32) ~[com.ning.async-http-client-1.9.21.jar:na]
        at com.ning.http.client.uri.Uri.create(Uri.java:25) ~[com.ning.async-http-client-1.9.21.jar:na]
        at com.ning.http.client.RequestBuilderBase.setUrl(RequestBuilderBase.java:307) ~[com.ning.async-http-client-1.9.21.jar:na]
        at com.ning.http.client.RequestBuilder.setUrl(RequestBuilder.java:165) ~[com.ning.async-http-client-1.9.21.jar:na]
        at play.api.libs.ws.ning.NingWSRequest.buildRequest(NingWS.scala:218) ~[com.typesafe.play.play-ws_2.10-2.4.6.jar:2.4.6]
        at play.api.libs.ws.ning.NingWSRequest.execute(NingWS.scala:128) ~[com.typesafe.play.play-ws_2.10-2.4.6.jar:2.4.6]
        at play.api.libs.ws.WSRequest$class.get(WS.scala:408) ~[com.typesafe.play.play-ws_2.10-2.4.6.jar:2.4.6]
        at play.api.libs.ws.ning.NingWSRequest.get(NingWS.scala:81) ~[com.typesafe.play.play-ws_2.10-2.4.6.jar:2.4.6]
        at de.frosner.broccoli.services.NomadService.de$frosner$broccoli$services$NomadService$$executeGetStatuses(NomadService.scala:38) ~[cluster-broccoli.cluster-broccoli-0.4.0-SNAPSHOT-sans-externalized.jar:na]
```

Clearly, this isn't very helpful and it's impossible to guess what the issue is.
Options should be sanitised and the app should log a useful error message and refuse to start.

FRosner:
Thanks for reporting this @omame!

PR #278, opened 2017-08-17
WIP: Download nomad logs from webui

swsnr:
Add buttons to download logs from the webui:

The download link gets the last 500 KiB of the log; the `offset` query parameter lets you go back further. 

Builds upon #277 which needs to be merged before this one.

### DIFF
--- a/server/src/it/scala/de/frosner/broccoli/nomad/NomadHttpClientIntegrationSpec.scala
+++ b/server/src/it/scala/de/frosner/broccoli/nomad/NomadHttpClientIntegrationSpec.scala
@@ -1,5 +1,8 @@
 package de.frosner.broccoli.nomad
 
+import com.netaporter.uri.Uri
+import com.netaporter.uri.dsl._
+import de.frosner.broccoli.nomad.models.Job
 import de.frosner.broccoli.test.contexts.WSClientContext
 import de.frosner.broccoli.test.contexts.docker.BroccoliDockerContext
 import de.frosner.broccoli.test.contexts.docker.BroccoliTestService.{Broccoli, Nomad}
@@ -25,19 +28,18 @@ class NomadHttpClientIntegrationSpec
   override def broccoliDockerConfig: BroccoliDockerContext.Configuration =
     BroccoliDockerContext.Configuration.services(Broccoli, Nomad)
 
-  private val baseUrl = "http://localhost:4646"
-  private val broccoliURL = "http://localhost:9000"
+  private val broccoliApi = "http://localhost:9000/api/v1"
 
   override def is(implicit executionEnv: ExecutionEnv): Any =
     "The NomadHttpClient" should {
       "get allocations for a running nomad job" >> { wsClient: WSClient =>
         // Generate a random identifier for the instance
         val identifier = Gen.resize(10, Gen.identifier).sample.get
-        val client = new NomadHttpClient(baseUrl, wsClient)
+        val client = new NomadHttpClient(Uri.parse("http://localhost:4646"), wsClient)
         (for {
           // Create and start a simple instance to look at it's allocations
           _ <- wsClient
-            .url(s"$broccoliURL/api/v1/instances")
+            .url(broccoliApi / "instances")
             .post(
               Json.obj("templateId" -> "http-server",
                        "parameters" -> Json.obj(
@@ -49,7 +51,7 @@ class NomadHttpClientIntegrationSpec
               response
             })
           _ <- wsClient
-            .url(s"$broccoliURL/api/v1/instances/$identifier")
+            .url(broccoliApi / "instances" / identifier)
             .post(Json.obj("status" -> "running"))
             .map(response => {
               response.status must beEqualTo(200)
@@ -57,7 +59,7 @@ class NomadHttpClientIntegrationSpec
               blocking(Thread.sleep(1.seconds.toMillis))
               response
             })
-          allocations <- client.getAllocationsForJob(identifier)
+          allocations <- client.getAllocationsForJob(shapeless.tag[Job.Id](identifier))
         } yield {
           (allocations.jobId === identifier) and (allocations.payload must have length 1)
         }).await(5, broccoliDockerConfig.startupPatience + 2.seconds)
--- a/server/src/main/resources/routes
+++ b/server/src/main/resources/routes
@@ -3,7 +3,10 @@
 # ~~~~
 
 # API routes
--> /api/v1  de.frosner.broccoli.ApiV1Router
+-> /api/v1  de.frosner.broccoli.routes.ApiV1Router
+
+# Downloads
+-> /downloads/ de.frosner.broccoli.routes.DownloadsRouter
 
 # Web-UI
 GET     /       @controllers.Assets.at(path="/public", file="index.html")
--- a/server/src/main/resources/routes
+++ b/server/src/main/resources/routes
@@ -2,30 +2,8 @@
 # This file defines all application routes (Higher priority routes first)
 # ~~~~
 
-# Templates
-GET     /api/v1/templates      @de.frosner.broccoli.controllers.TemplateController.list
-GET     /api/v1/templates/:id  @de.frosner.broccoli.controllers.TemplateController.show(id: String)
-
-# Instances
-GET     /api/v1/instances         @de.frosner.broccoli.controllers.InstanceController.list(templateId: Option[String])
-GET     /api/v1/instances/:id     @de.frosner.broccoli.controllers.InstanceController.show(id: String)
-POST    /api/v1/instances         @de.frosner.broccoli.controllers.InstanceController.create
-POST    /api/v1/instances/:id     @de.frosner.broccoli.controllers.InstanceController.update(id: String)
-DELETE  /api/v1/instances/:id     @de.frosner.broccoli.controllers.InstanceController.delete(id: String)
-GET  /api/v1/instances/:id/tasks  @de.frosner.broccoli.controllers.InstanceController.tasks(id: String)
-
-
-# About
-GET     /api/v1/about  @de.frosner.broccoli.controllers.AboutController.about
-
-# Status
-
-GET     /api/v1/status  @de.frosner.broccoli.controllers.StatusController.status
-
-# Authentication
-POST    /api/v1/auth/login   @de.frosner.broccoli.controllers.SecurityController.login()
-POST    /api/v1/auth/logout  @de.frosner.broccoli.controllers.SecurityController.logout()
-GET     /api/v1/auth/verify  @de.frosner.broccoli.controllers.SecurityController.verify()
+# API routes
+-> /api/v1  de.frosner.broccoli.ApiV1Router
 
 # Web-UI
 GET     /       @controllers.Assets.at(path="/public", file="index.html")
--- a/server/src/main/scala/de/frosner/broccoli/nomad/NomadModule.scala
+++ b/server/src/main/scala/de/frosner/broccoli/nomad/NomadModule.scala
@@ -3,6 +3,7 @@ package de.frosner.broccoli.nomad
 import javax.inject.Singleton
 
 import com.google.inject.{AbstractModule, Provides}
+import com.netaporter.uri.Uri
 import net.codingwell.scalaguice.ScalaModule
 import play.api.Configuration
 import play.api.libs.ws.WSClient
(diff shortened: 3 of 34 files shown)


server/src/it/scala/de/frosner/broccoli/nomad/NomadHttpClientIntegrationSpec.scala
```diff
@@ -49,15 +51,15 @@ class NomadHttpClientIntegrationSpec
...
+            .url(broccoliApi / "instances" / identifier)
             .post(Json.obj("status" -> "running"))
             .map(response => {
               response.status must beEqualTo(200)
               // Wait until the service is up
               blocking(Thread.sleep(1.seconds.toMillis))
               response
             })
-          allocations <- client.getAllocationsForJob(identifier)
+          allocations <- client.getAllocationsForJob(shapeless.tag[Job.Id](identifier))
```
FRosner, line 62: Thanks for introducing tagges types!

server/src/main/scala/de/frosner/broccoli/nomad/models/Allocation.scala
```diff
@@ -2,20 +2,31 @@ package de.frosner.broccoli.nomad.models
...
 /**
   * A partial model for a single allocations.
   */
-final case class Allocation(id: String,
-                            nodeId: String,
-                            clientStatus: ClientStatus,
-                            taskStates: Map[String, TaskStateEvents])
+final case class Allocation(
+    id: String @@ Allocation.Id,
+    jobId: String @@ Job.Id,
```
FRosner, line 13: Shall we create a type alias for that?

`type JobId = String @@ Job.Id`?

server/src/main/resources/routes
```diff
@@ -3,7 +3,10 @@
 # ~~~~
 
 # API routes
--> /api/v1  de.frosner.broccoli.ApiV1Router
+-> /api/v1  de.frosner.broccoli.routes.ApiV1Router
+
+# Downloads
+-> /downloads/ de.frosner.broccoli.routes.DownloadsRouter
```
FRosner: What is the reasoning behind not making this part of our versioned API? IMHO downloading logs for an instance task allocation is not different from asking for its parameters. Also `/downloads` sounds like its a folder mapping where you can download static files. But it's actually something to refresh to get more logs.

@Gerrrr @lunaryorn what do you think?

server/src/main/scala/de/frosner/broccoli/nomad/NomadModule.scala
```diff
@@ -37,5 +38,5 @@ class NomadModule extends AbstractModule with ScalaModule {
   @Provides
   @Singleton
   def provideNomadClient(config: NomadConfiguration, wsClient: WSClient): NomadClient =
-    new NomadHttpClient(config.url, wsClient)(play.api.libs.concurrent.Execution.defaultContext)
+    new NomadHttpClient(Uri.parse(config.url), wsClient)(play.api.libs.concurrent.Execution.defaultContext)
```
FRosner, line 41: Does this change resolve the problem reported in https://github.com/FRosner/cluster-broccoli/issues/81?

FRosner (changes requested):
Looks really nice! Thanks a lot. I assume that after rebasing I anyway will have to review again so let's discuss the few comments I had before merging. I also wanted to look at the UI part and maybe make a few small changes to the layout.

server/src/main/scala/de/frosner/broccoli/nomad/NomadModule.scala
swsnr: Perhaps… I don't know what message `Uri.parse` throws.

swsnr:
@FRosner What are these changes?  Why did you push them to this branch?

Wasn't too nice, to be honest… the pull request was still WIP, and not ready.  Now I've got to rebase your changes on top of my local changes 😒

webui/src/Views/InstanceView.elm
```diff
@@ -45,6 +45,14 @@ jobControlsColumnWidth =
     200
 
 
+allocationStateColumnWidth =
+    190
+
+
+allocationLogColumnWidth =
+    110
```
swsnr: Why do we need fixed widths for columns?

webui/src/Views/InstanceView.elm
```diff
@@ -412,7 +455,29 @@ instanceAllocationRow index ( taskName, allocation ) =
...
             , td [] [ code [] [ text (shortAllocationId allocation.id) ] ]
-            , td [] [ span [ class ("label " ++ labelKind) ] [ text description ] ]
+            , td
+                [ width allocationStateColumnWidth
+                , class "text-center"
+                ]
+                [ span [ class ("label " ++ labelKind) ] [ text description ] ]
+            , td
+                [ width allocationLogColumnWidth
+                , class "text-center"
```
swsnr: Why center?

FRosner:
@lunaryorn the changes are related to the UI part of this change. I pushed them to this branch in order to support you. I understand that it is WIP but you requested my review so that's what I did. During the review I found something I didn't like so I changed it. If you struggle with rebasing you can also cherry-pick?

webui/src/Views/InstanceView.elm
FRosner: It looks better.

webui/src/Views/InstanceView.elm
FRosner: If a column has fixed content I prefer to have the width fixed. This way the layout is more consistent across multiple of those tables.

swsnr:
I've force-pushed to fix the test, use a different colours for the buttons (namely `info`) and use fixed table layout for the tasks table to make sure that all those tables have the same column widths:

bildschirmfoto 2017-08-18 um 14 33 39

swsnr:
@FRosner In contrast to fixed column width this approach equally distributes whitespace in the table.  Large whitespace between allocation and state makes it hard to scan the table, imho, as the state's visually disconnected from the allocation cell.

swsnr:
Smaller display:

bildschirmfoto 2017-08-18 um 14 38 10

FRosner:
When you look at this table, I find it quite visually imbalanced. As discussed, let's reorder the table, putting allocation id and state in the first two columns, placing the task name in the middle claiming as much space as possible, and aligning the log buttons to the right.

image

The button color should not be more offensive than the start and stop buttons, IMHO. I liked them white, like all the other buttons.

Also I think that "Logs" describes the column good enough, because technically these are not files. They are log streams and you get a snapshot when you click on it. When you refresh, you get a new snapshot. What bothers you with "Logs"?

swsnr:
How about this?

bildschirmfoto 2017-08-21 um 10 14 30

server/src/main/scala/de/frosner/broccoli/nomad/models/Allocation.scala
swsnr: Personally I don't need them, I find they introduce a lot of clutter and boilerplate without any significant benefit.  These types are short; aliasing them won't save much will it?

server/src/main/scala/de/frosner/broccoli/nomad/NomadModule.scala
swsnr: IOW, I didn't make this change to address #81, but just to build URLs in the client in a simpler way, and I don't think that it'll result in better error messages.

To address #81 you'd need to validate the URL upfront while parsing the configuration, to be able to throw an error message that can still refer to the path of the malformed value.

swsnr:
@FRosner Ping :)

FRosner:
Sorry :D

I'm looking right now.

FRosner (approved):

FRosner:
Awesome!

server/src/main/resources/routes
Gerrrr: Yep, I think that it should be at `/api/v1` since it's a part of the API.

swsnr: I don't think that this is an API endpoint.  

Can I send it JSON? Does it return JSON?  Do we use it as an API endpoint?  I don't think so: We use it as a static endpoint for the webui, and it's not intended to serve structured data but rather human-readable "logs" whatever they might contain.

As for `/downloads/`: It needs to be in a separate URL prefix to allow proper 404s for non-existing logs, and I couldn't come up with a better prefix.  While I agree that it's not a "static file" it's not a "dynamic view" either: The log won't refresh, so in a way you're really "downloading" a snapshot of the log at this particular point of time.

FRosner: It does not send JSON and it does not receive JSON. It receives a GET request and returns a text. But the format does not determine whether it is part of an API or not. For me there are two types of endpoints:

1. Endpoints that are meant to be accessed by a browser, serving a webpage with a layout, etc. This is the index.html basically. I'd call those part of the UI.
2. Endpoints that are meant to be accessed by programs, CLIs, scripts. This is useful for interacting with Broccoli without using a browser. I'd call those part of the API.

While we are certainly using the endpoint through a browser this is just temporary. We do it because we didn't include the logs directly in the UI. But what you get there is a resource and the protocol should be versioned. Why do I care? If someone uses Broccoli from a script or from the command line and we change the API (e.g. no longer return text but JSON), then it should be reflected in a new version of the API.

I agree that not sending the logs in JSON is a bit inconsistent. But we can make that call. We just need to document it.

I didn't fully get the separate URL prefix issue. Each endpoint can return 404 individually. If you request an instance that does not exist, you get 404. Same for logs.

swsnr: > While we are certainly using the endpoint through a browser this is just temporary.

If it's just temporary what's going to replace it? What will replace this endpoint to provide the ability to, well, "download" a snapshot of the log to view (grep, less, etc.) it on the local system?

I'm aware that we'll stream logs at some point but is that a replacement for the ability to get the log on the local system?  I can't `grep` in the webui, can I?

> But what you get there is a resource and the protocol should be versioned.  Why do I care? If someone uses Broccoli from a script or from the command line and we change the API […]

The resource offers logs as they are on nomad node.  How could this endpoint change?  Or, in other words, if we wanted to change this endpoint wouldn't that rather be a change in the semantics and thus be a wholly new endpoint?

> I didn't fully get the separate URL prefix issue. Each endpoint can return 404 individually. If you request an instance that does not exist, you get 404. Same for logs.

Try `curl localhost:9000/this-does-not-exist` versus `curl localhost:9000/downloads/this-does-not-exist`:  The latter returns an empty 404, whereas the former returns a HTML page.

Since we already have a top-level catch-all route (for assets) that's only possible with a separate prefix.  

For the same reason there's a dedicated API router now.

FRosner: The endpoint can stay as it is very useful. What I'm saying is that it is not part of the UI. And anything that is not part of the UI, is part of the API, in my opinion. Then it makes more sense to add it as part of the API from a routing point of view, no?

I don't know how this endpoint is going to evolve but I see advantages of having a stable API and versioning it at the same time. This does not conflict IMO.

I didn't get the last part with the curls. If the dedicated API router is there why doesn't it return 404 if the API route does not exist?

@Gerrrr what do you think?

swsnr: > What I'm saying is that it is not part of the UI. 

The UI uses it as of now.

If we moved it to the stable and versioned API we would have to provide more guarantees than we do currently.  We would need to introduce content negotiation to support JSON and plain text, we would have to specify parameter ranges for the log size, etc.

All of this is missing currently, because I treated it as a UI endpoint.

As an example of the implication of "stabilizing" this endpoint:  What happens if the user tries to load a 2GB log file?  Currently we allocate a string for the entire logfile (because that's simple), but since the UI only gets the last some KBs there's no harm.  If we stabilize the endpoint we need to refactor the code to stream the log file from Nomad to the frontend to avoid excessive memory allocation for large log files.  That's easy for plain text, but I don't know whether Play can encode large strings in a streaming fashion. 

If it can't we suddenly need to two code paths for different content types or impose a maximum limit on the log size, but then how to we pick this limit?  How to we make sure the limit cuts the log file correctly, ie doesn't split in the middle of surrogate pairs rendering the log unreadable to the client?

If we can provide satisfying answers to these questions we can make it part of the stable API.  If we can't we shouldn't, because—since it is a "stable" API—any bug we make becomes part of the interface and can't be fixed without introducing new API versions or potentially breaking clients.

> I don't know how this endpoint is going to evolve.

Then why plan for change management?    Why not keep it as UI endpoint that lets us change it if *we* need it, without consideration for compatibility?

Don't be too fast with declaring something "stable" without giving it due time to ripe :wink:

> I didn't get the last part with the curls. If the dedicated API router is there why doesn't it return 404 if the API route does not exist?

It does, but I think we're talking past each other.  I referred to why I introduced `/downloads` as a prefix instead of adding a top-level route, not about `/downloads/` versus `/api/`.

As I said I do not think that this endpoint is part of the API.

FRosner: Ok. To summarize we don't have an API endpoint to get logs at the moment, only a very ugly part of the UI. Correct?

Works for me! 👍

swsnr: @FRosner Yes that's the situation from my point of view.  From this state we can slowly refine to  get a nicer UI… an API for logs is less important to me as long as we don't have specific need.

FRosner: Agreed thank you!

PR #279, opened 2017-08-17
Remove logging trait

swsnr:
Get rid of the Logging trait which was just a complicated and obscure way for `val log = Logger(getClass)`.  

Do the latter instead of mixing in the Logging trait in all classes that log.

Replace the pseudo-logger for execution times with a simple decorator function that takes a block and a logger function, measures the block and sends the message to the given logging function.

All in all, less inheritance, less classes, easier code.


Gerrrr (approved):

FRosner (approved):
LGTM. Just for me to understand, what is the problem with the Logging Trait? What's the difference to, e.g., the `LazyLogging` or `StrictLogging` traits provided by [scala-logging](https://github.com/typesafehub/scala-logging)?

swsnr:
@FRosner Let me return the question: What purpose did the logging trait serve?  What did the trait abstract over, and was that abstracting worth paying the price, ie, leaking "logging" into the inheritance hierarchy and the public interface of classes?

----

Wrt to scala-logging, the `Logging` trait wasn't much different, but these traits exist to model logging in a heavily mixin-based architecture, eg, a "traditional" cake-pattern based code base.  You'd use these traits to allow mixins access the logger of the target class and enforce a common naming, so as not to end up with different logger conventions in different mixins, ie, one mixin wanting "log" and the other "logger", etc.

But we don't have a lot of mixins here, do we?

----

We shouldn't use mixins a lot either, as Play already has an infrastructure for dependency injection, namely Guice, and arguably mixins cause a lot of trouble when used at scale—the Cake pattern isn't considered a success these days, and typically avoided in contemporary scala applications.

FRosner:
Thanks for the explanation @lunaryorn. Why don't we dependency inject the logger?

swsnr:
@FRosner Well, the logger should be named after the class that's using it, shouldn't it?  How'd you do that with dependency injection?

And then, why should we inject the logger?  What problem would that solve?  The logger's no issue in tests, and would we ever want to change it?

FRosner:
Ok thank you.

PR #281, opened 2017-08-21
Move docker production build to SBT

swsnr:
Removes `docker/prod` and builds the docker production image with sbt-native-packager instead.

The docker production build now comes down to `docker login; sbt docker:publish` and does not require `prepare-docker-builds` anymore.  SBT completely hides temporary artifacts of the docker build (no more unzipping and copying around) and automatically adds the binary distribution to the docker image.

To inspect the temporary artifacts, run `sbt docker:stage` and inspect `server/target/docker/stage`.  To build the image in the local docker daemon, without publishing to hub, run `sbt docker:publishLocal`.

## Notable differences to the old image

The new image looks like follows

```
FROM openjdk:8-jre
MAINTAINER Frank Rosner
ADD templates /templates
RUN ["mkdir", "/instances"]
RUN ["chown", "-R", "daemon:daemon", "/templates", "/instances"]
RUN ["chmod", "0755", "/instances"]
WORKDIR /opt/docker
ADD opt /opt
RUN ["chown", "-R", "daemon:daemon", "."]
EXPOSE 9000
USER daemon
ENTRYPOINT ["bin/cluster-broccoli"]
CMD ["-Dbroccoli.templates.storage.fs.url=/temlates", "-Dbroccoli.instances.storage.fs.url=/instances"]
```

* It has a `MAINTAINER` tag which supposedly is required for docker hub.
* It runs Broccoli as daemon user, with limited privileges.
* Broccoli is at `/opt/docker` instead `/cluster-broccoli` (that'd be configurable).  Instances are `/opt/docker/instances`, templates at `/opt/docker/templates` by default.
* The image has a `WORKDIR`, ie, you can start the application without specifying the full path as in `bin/cluster-broccoli` as opposed to `/cluster-broccoli/bin/cluster-broccoli`.

## Notable differences in the Travis CI build

* Image tags for branches and travis build numbers are gone.  I could add them back, but do we actually use these tags anywhere?  We still tag with `latest` and the short commit ID.

PR #282, opened 2017-08-21
Remove comment about dev docker image

FRosner:
It was deleted already as it was unused


swsnr:
@FRosner I fixed that in #281

PR #283, opened 2017-08-22
Run Elm tests with "sbt webui/test"

swsnr:


.travis.yml
```diff
@@ -19,19 +19,19 @@ jobs:
...
-        - cd webui
-        - yarn install
-        - yarn setup
       script:
-        - yarn format:validate
-        - yarn compile
-        - yarn test
+        - sbt webui/yarnFormatValidate
+        # Generate resources to ensure that webpack builds cleanly
+        - sbt webui/copyResources
```
FRosner, line 32: What does copyResources do exactly and when is it necessary?

FRosner (changes requested):
Thanks for the change!

We cannot run yarn start from there, right? So for this we would still navigate to `webui` and execute `yarn start`?

Let's mention the new commands somewhere? Is it still useful to use yarn directly? If not, should we change the information in the `webui/README.md` file?

.travis.yml
swsnr: It copies all resources into the binary output directory, ie, `/target/scala-X.XX/classes/`, which implies running all resource generators, ie, webpack in our case.  In the end this triggers a webpack build from SBT.

It's the only way to trigger resources generators, as there's no dedicated command for this.  Note that `sbt compile` does not generate resources, because resources are not strictly required for compilation.  They only trigger when compiling across project dependencies, ie, `server/compile` will trigger resources generators of webui.

---

Try running `sbt webui/clean webui/copyResources`.  You'll see that SBT starts a webpack build after which the generated assets end up in `webui/target/scala-2.11/classes`.  This directory then ends up in classpath for `server/run`, and gets jar'ed up when building `sbt dist`.

swsnr:
Nothing changed for `yarn` in this PR.  As far as I'm concerned it's perfectly fine to continue using `yarn`—in fact I find it more convenient when working on Elm only.

But when working on the webui and the server at the same time it's also very convenient to be able to run `sbt test` and really run the entire test suite of the application, across both subprojects, right from Intellij's SBT shell.

swsnr:
I wouldn't document anything.  In the end this just makes `sbt webui/test` do what it's really supposed to do: Run tests in the webui project :wink:

I changed `.travis.yml` just because SBT makes it shorter: No changing directories—I always found it weird that `cd` in `install` gets preserved for `script` 😕—and `yarn install`/`yarn setup` run automatically due to dependencies.

Also `sbt dist` is what defines our production build, so using `sbt` for the frontend build makes it closer to the actual production build.

.travis.yml
FRosner: So is this step automatically executed when I run `sbt dist`?

FRosner (approved):

PR #288, opened 2017-08-24
#17 Escape string parameters instead of converting them to JsString

Gerrrr:


server/src/test/scala/de/frosner/broccoli/templates/TemplateRendererSpec.scala
```diff
@@ -52,14 +52,14 @@ class TemplateRendererSpec extends Specification with Mockito {
         id = "1",
         template = Template(
           id = "1",
-          template = "{{id}}",
+          template = """"{{id}}"""",
```
FRosner, line 55: """"ok""""

FRosner (approved):

build.sbt
```diff
@@ -20,12 +20,12 @@ lazy val server = project
       Dependencies.scalaUri,
       Dependencies.squants,
       Dependencies.scalaguice,
+      Dependencies.commonsIO,
```
swsnr, line 23: The feature you need is from commons lang.  I think that's a separate dependency.

If it is please use commons labg here and keep commons IO as test dependency.

Relying on transitive dependencies is risky.

swsnr:
What's the purpose of this change?

build.sbt
Gerrrr: Thanks for the catch! You are right about commons lang. I'm going to fix it in a separate PR.

Gerrrr:
In #265 all string parameter values are rendered in templates as valid JSON strings - escaped and wrapped by quotes. The issue is that we do not always want to have the quotes directly around the parameter value. Example: [a service name that consist of id and a constant string](https://github.com/FRosner/cluster-broccoli/blob/master/templates/http-server/template.json#L35). If `id` parameter is of string type, the rendered template would look like this - `""service-name"-web-ui-1"`. With the previous implementation it took a lot of effort in the template logic to get rid of the extra pair of quotes around the parameter.

We discussed with @FRosner that it is more convenient and simpler to use if the string type would just make sure that the parameter is an escaped string and it is the responsibility of a template creator to put quotes where they are needed.

swsnr:
@Gerrrr I see; thanks for explaining.

Issue #287, opened 2017-08-23
Resource Allocation Information


FRosner:
### Problem

When configuring the CPU, disk and memory of your instances it is often a bit of a guesstimate.

### Solution

For each task, show the resource information, like it is done through `nomad alloc-status`. We should start with visualizing CPU and memory for now.

```
Task "broccoli" is "running"
Task Resources
CPU          Memory           Disk  IOPS
826/500 MHz  595 MiB/2.0 GiB  0 B   0
```

#### Branch

https://github.com/FRosner/cluster-broccoli/tree/frosner/issue-287-resource-allocation-information-ui-draft

#### Overview

![image](https://user-images.githubusercontent.com/3427394/29625550-5143704e-882c-11e7-900f-5400c3d97576.png)

#### Hover Title

![untitled 2](https://user-images.githubusercontent.com/3427394/29625887-56b0f35c-882d-11e7-89ca-3140df91ee4a.png)

#### Showing the Percentage

![image](https://user-images.githubusercontent.com/3427394/29626062-dcca3f02-882d-11e7-95e5-258983a5f5dd.png)

#### Color Coding (Green, Yellow, Red, Red Striped)

![image](https://user-images.githubusercontent.com/3427394/29626228-5a5aa114-882e-11e7-8107-2f1375257f42.png)

### Open Points

- As we need to request this information from each individual nomad client we probably don't want to do this very often. Actually it would be optimal to do it lazily (so only when someone expands an instance) and maybe refresh it every 5 seconds or something. It doesn't have to be too interactive as we can use a more sophisticated infrastructure involving a metrics DB etc. for that. So it should be an indication whether your planned resources are planned well.
- It would be great to have an intuitive visualization and then show the actual numbers only when hovering. We could use a horizontal bar (like in a progress bar). We only need to have a special encoding for values that exceed the visual maximum (like 800 / 500 MHz CPU).

swsnr:
I'll take a look next week to figure out how to get the resources from the Nomad API.

swsnr:
@FRosner Which resources are we interested in?

I presume CPU and memory, but what about disk?  And network speed?  I can't find the latter two in the corresponding Nomad API…

FRosner:
Let's start with CPU and memory. AFAIK Nomad does not really limit disk, @Gerrrr? Also we don't have any use case where disk plays a role at the moment. Network is not configurable right now.

It is certainly possible that we add it later but right now we don't need it.

[...]

PR #293, opened 2017-08-29
Show CPU & memory usage in webui

swsnr:
For #287 

Shows the raw numbers currently.  Comparing resources against the reserved resources by the Nomad job description as suggested in #287 comes next, but not as part of this PR.


webui/src/Views/InstanceView.elm
```diff
@@ -441,23 +437,28 @@ instanceAllocationRow instance index ( taskName, allocation ) =
...
         tr []
-            [ td [] [ code [] [ text (shortAllocationId allocation.id) ] ]
+            [ td [] [ code [] [ text (shortAllocationId task.allocationId) ] ]
             , td [ class "text-center" ]
                 [ span [ class ("label " ++ labelKind) ] [ text description ]
                 ]
-            , td [] [ text taskName ]
+            , td [] [ text task.taskName ]
+            , td [ class "text-center" ]
+                [ text (unwrap "unknown" (Round.round 2) task.cpuTicksUsed)
```
FRosner: We should also add the unit here (MHz).

webui/webpack.config.js
```diff
@@ -83,7 +83,7 @@ module.exports = {
         test: /\.elm$/,
         exclude: [/elm-stuff/, /node-modules/],
         use: [
-          { loader: 'elm-hot-loader' },
+          // { loader: 'elm-hot-loader' },
```
FRosner: Why did you comment out the hot loader?

server/src/main/scala/de/frosner/broccoli/websocket/WebSocketMessageHandler.scala
```diff
@@ -58,3 +63,53 @@ class BroccoliMessageHandler @Inject()(instances: NomadInstances, instanceServic
...
+  *
+  * @param underlying The underlying message handler.
+  */
+class CachedBroccoliMessageHandler @Inject()(
+    underlying: BroccoliMessageHandler,
+    cache: CacheApi
+)(implicit ec: ExecutionContext)
+    extends WebSocketMessageHandler {
+
+  private val timeout: Duration = 3.seconds
```
FRosner: This is how I understand the mechanism: Instead of limiting the clients in the amount of requests they can make to Broccoli but then forward every request to nomad directly we are allowing the clients to make as many requests as they like, but then cache the nomad response for three seconds. Is this correct?

I think we should make the cache timeout configurable so we can find an optimal tradeoff between the up-to-dateness and the load we put on Nomad. What do you think?

webui/webpack.config.js
swsnr: It broke with some elm module that I introduced.  I'm not sure why, but I wanted to take a look first.

webui/src/Views/InstanceView.elm
swsnr: I'm not sure these are MHz.  Nomad exposes them as CPU ticks, in my understanding that's not the same as MHz.

server/src/main/scala/de/frosner/broccoli/websocket/WebSocketMessageHandler.scala
swsnr: That's the approach, yes.  I'll add a configuration value for the timeout.

webui/src/Views/InstanceView.elm
FRosner: It corresponds to CPU shares (MHz) and we are going to put it into relation to the allocated CPU shares (MHz) in order to get the percentage.

```
Task Resources
CPU         Memory          Disk     IOPS  Addresses
33/100 MHz  18 MiB/128 MiB  300 MiB  0     ui: 127.0.0.1:8000
```

image

- https://www.nomadproject.io/docs/agent/telemetry.html#host-metrics

swsnr: @FRosner I was confused by the name “ticks“ to be honest.  In Unix "tick" normally refers to the internal timer of the kernel not to CPU clock speed.

swsnr:
@FRosner Can you review again?  This part is ready for merge as far as I'm concerned.

FRosner:
Will do!
-- 
Sent from my phone. Please excuse my brevity.

server/src/main/resources/reference.conf
```diff
@@ -41,6 +41,12 @@ broccoli {
     path = "./templates"
   }
 
+  # Configuration for Broccoli's Websocket connection with the webui
+  web-socket {
+    # How long to cache responses to websocket requests in Broccoli
+    cache-timeout = 3 seconds
```
FRosner, line 47: We could actually use this timeout also for caching the other requests (like service and job status) later on, right? So the good design would be to ask Nomad on demand (not every x seconds) for the job status and everything but then cache the results?

FRosner:
LGTM. Just one comment: Right now the column size for CPU and Memory is not fixed. If the metrics now change, the columns will become bigger and smaller. I don't think this is an issue, because as soon as we replace the numbers by fixed width bars we shouldn't have this problem anymore.

FRosner (approved):

server/src/main/resources/reference.conf
swsnr: @FRosner Yes, that's what I'd do.

PR #295, opened 2017-09-04
Add silhouette

swsnr:
Add some basic silhouette support, by making use of some Silhouette types:

- Use Silhouette's `Credentials` type for, well, credentials
- Use Silhouette's `IdentityService` to resolve user names to accounts
- Use Silhouette's `CredentialsProvider` together with `AuthInfoRepository` to authenticate users.

This provides a huge part of the infrastructure requires to use Silhouette's request authentication.  Essentially all that remains is provide a `SilhouetteProvider` in `AuthModule`, then have `Silhouette` injected into controllers that authenticate and use it for authentication.

## Breaking changes

Moved all auth configuration into a case class, affecting the following keys:

* `broccoli.auth.session.allowMultiLogin` → `broccoli.auth.session.allow-multi-login`
* `broccoli.auth.allowedFailedLogins` → `broccoli.auth.allowed-failed-logins`
* `broccol.auth.conf.accounts[].instanceRegex` → `broccol.auth.conf.accounts[].instance-regex`

Hence the `conf` object is almost empty now, meaning that almost all settings are now defined and documented in `reference.conf`, and loaded through pureconfig into the super-config-case-class.

## Misc changes

Simplified the `Account` stuff.  The separate trait is gone, `Anonymous` is now simply a value in the companion object.

Also `Account` has no `password` anymore.  Silhouette strictly separates between `Identity`, ie, username, permissions, what we normally call a "user profile", and authentication information, ie, a hashed password.  

This enables Silhouette to attach multiple ways to authenticate, like Facebook auth, username & password, etc. to a single identity, so that a user can login in different ways.  We don't need this, but we can still separate public from confidential data, ie, passwords on the type level this way.

Passwords are only in the `AuthInfoRepository` now, which is setup in `AuthModule` and exposes through the `CredentialsProvider`, which takes user credentials and checks them against all known authentication information.

As part of this change passwords are now hashed with bcrypt when they are added to the authentication info repo in `AuthModule`.  Silhouette makes this trivial, and it slows down authentication, making brute-force attacks against passwords less feasible.


FRosner:
@lunaryorn as far as I remember this PR was already working as is? So I can merge it and then work, e.g., on the username and password authentication based on LDAP, right?

Thanks!

swsnr:
@FRosner I do not remember.  I think it's self-contained, but you'll have to test.

PR #296, opened 2017-09-07
don't show an error message to the client if allocation stats cannot be parsed

FRosner:
When restarting an instance the allocation stats are not available for
a short period of time. If we request them at this moment the result
will not be a valid JSON but also not a connect error.


FRosner:
@lunaryorn can you please take a look?

Gerrrr (dismissed):

PR #310, opened 2017-10-27
Reset model also if there is a WSConnectError

### LINKED ISSUE #219: Logged in but I see the login form
grafik

FRosner:
fixes #219

### DIFF
--- a/webui/src/Updates/UpdateLoginStatus.elm
+++ b/webui/src/Updates/UpdateLoginStatus.elm
@@ -6,7 +6,6 @@ import Model exposing (Model)
 import Models.Ui.LoginForm as LoginForm exposing (LoginForm)
 import Utils.CmdUtils as CmdUtils
 import Ws
-import Time
 import Debug
 import Routing
 import Http exposing (Error(..))
@@ -71,17 +70,17 @@ updateLoginStatus message model =
                         case response.status.code of
                             403 ->
                                 ( { model | authRequired = Just True }
-                                , CmdUtils.delayMsg (5 * Time.second) AttemptReconnect
+                                , Ws.attemptReconnect
                                 )
 
                             _ ->
                                 ( model
-                                , CmdUtils.delayMsg (5 * Time.second) AttemptReconnect
+                                , Ws.attemptReconnect
                                 )
 
                     _ ->
                         ( model
-                        , CmdUtils.delayMsg (5 * Time.second) AttemptReconnect
+                        , Ws.attemptReconnect
                         )
             else
                 ( model
--- a/webui/src/Updates/UpdateLoginStatus.elm
+++ b/webui/src/Updates/UpdateLoginStatus.elm
@@ -65,20 +65,25 @@ updateLoginStatus message model =
                 )
 
         FetchVerify (Err error) ->
-            case error of
-                BadStatus response ->
-                    case response.status.code of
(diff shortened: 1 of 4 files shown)


FRosner:
@Gerrrr can you take a look? Please also note the discussion in the issue.

webui/src/Updates/UpdateLoginStatus.elm
```diff
@@ -65,20 +65,25 @@ updateLoginStatus message model =
...
+                            403 ->
+                                ( { model | authRequired = Just True }
+                                , CmdUtils.delayMsg (5 * Time.second) AttemptReconnect
+                                )
 
-                        _ ->
-                            ( model
-                            , CmdUtils.delayMsg (5 * Time.second) AttemptReconnect
-                            )
+                            _ ->
```
Gerrrr, line 76: mixed spaces and tabs

webui/src/Updates/UpdateLoginStatus.elm
```diff
@@ -65,20 +65,25 @@ updateLoginStatus message model =
...
+                                ( model
+                                , CmdUtils.delayMsg (5 * Time.second) AttemptReconnect
+                                )
 
-                _ ->
-                    ( model
-                    , CmdUtils.delayMsg (5 * Time.second) AttemptReconnect
-                    )
+                    _ ->
```
Gerrrr, line 81: Is it possible to combine type checking of the `response` and the value comparision of its attribute `status.code`? This would avoid code duplication and make this snippet a bit more readable.

Gerrrr (changes requested):

webui/src/Updates/UpdateLoginStatus.elm
FRosner: We are using elm-format to format the code and it is also validated. WHat do you mean exactly?

webui/src/Updates/UpdateLoginStatus.elm
FRosner: I see your point. The problem is that I first need to deconstruct the `BadStatus` before I can get the response. I tried pattern matching the status code directly but it didn't work and I couldn't find any information online.

In Scala you can use [guards](https://alvinalexander.com/scala/how-to-use-if-then-expressions-guards-in-case-statements-scala) for this but Elm doesn't have this feature.

What I will do is at least factor our the reconnect command `CmdUtils.delayMsg (5 * Time.second) AttemptReconnect` into a separate function and use it everywhere.

Does that work for you?

Gerrrr: Yup, I also tried to find a way to pattern match the type and the attribute value, but did not succeed.

Sure, what you suggested makes sense.

Thank you!

webui/src/Updates/UpdateLoginStatus.elm
Gerrrr: I thought that those darker spaces are the actual spaces mixed with tabs. Looks like Github just shows this way that the arrow is moved to the right.

webui/src/Updates/UpdateLoginStatus.elm
FRosner: Done!

Gerrrr (approved):

PR #311, opened 2017-10-29
Introduce jinjava for template rendering

### LINKED ISSUE #224: More powerful template engine
### Problem

Right now we can only have simple string replacement. It would be cool to have a more powerful template engine allowing control structures likes conditions, loops, etc.

We only need to take care that there is a way to identify the variables required so we can still ask the user for the input accordingly.

### Solutions

Use [https://github.com/HubSpot/jinjava](Jinjava). The process would be the following:

- When creating a template, you can use Jinja2 syntax.
- When the templates are loaded, we cannot really render or check anything as Jinjava needs parameter values in order to render.
- As soon as the template is rendered (which is before submitting the instance job to Nomad), we take the parameter values specified by the user and feeding them to Jinjava for rendering.

Because Jinja allows for locally bound variables (e.g. in loops), we can no longer extract all possible variables from the template. So the UI does not know which parameters to display. Instead, when rendering the UI we have to rely on the `template.conf` information.

### VIAC (Very Important Acceptance Criterion)

- It is not possible to have an unbound variable in the Jinja template to which there is no information in the `template.conf` and then still bind it by submitting the parameter values including this variable. Instead, Broccoli should fail.

### ToDo

- [x] The Jinja template characters (`{{`, and `}}`) will conflict with the template stanza from Nomad 0.5.0. However the template characters from Nomad are [configurable](https://www.nomadproject.io/docs/job-specification/template.html#left_delimiter) so that should be fine. Can Jinjava also have configurable template characters?
- [x] The user specifies all possible parameters in the parameter info section of the `template.conf`. If those parameters are not enough to completely render the template, Jinjava should fail. It knows this because there are unbound variables. Can we fail on unbound variables?

---

# Documentation changes

## https://github.com/FRosner/cluster-broccoli/wiki/Upgrading-to-0.8.0

**Templates**

Templates are now rendered with [Jinjava template engine](https://github.com/HubSpot/jinjava). Because of this, there are several changes breaking backwards compatibility:
* Variables can no longer contain dashes. In order to automatically convert dashes in the variable names in instances and templates, you can set `broccoli.templates.convert-dashes-to-underscores` to `true`.
* Broccoli no longer automatically discovers template variables. You have to explicitly specify all variables in `template.conf`. To fix existing instances, please run the upgrade script:
    - `nomad stop cluster-broccoli`
    - backup instances directory
    - `python script/instances-0.7.0-to-0.8.0.py /path/to/broccoli/instances`
    - `nomad start cluster-broccoli`

**Configuration Changes**

It is possible to set [Jinjava config settings](https://static.javadoc.io/com.hubspot.jinjava/jinjava/2.4.0/index.html?com/hubspot/jinjava/JinjavaConfig.html) through Broccoli. The default values are kept the same as in Jinjava, unless stated otherwise. For documentation on those parameters please refer to the original project https://github.com/HubSpot/jinjava.

Here is the list of Jinjava settings configurable through Broccoli:

* `broccoli.templates.jinjava.trim-blocks` (defaults to `false`)
* `broccoli.templates.jinjava.lstrip-blocks` (defaults to `false`)
* `broccoli.templates.jinjava.max-render-depth` (defaults to `10`)
* `broccoli.templates.jinjava.read-only-resolver` (defaults to `true`)
* `broccoli.templates.jinjava.enable-recursive-macro-calls` (defaults to `false`)
* `broccoli.templates.jinjava.fail-on-unknown-tokens` (defaults to `true`)
* `broccoli.templates.jinjava.max-output-size` (defaults to `0`)
* `broccoli.templates.jinjava.nested-interpretation-enabled` (defaults to `true`)

---

### Nomad JSON Template File

The template file is a Jinja2 template of a [Nomad job file in JSON format](https://www.nomadproject.io/docs/jobspec/json.html). In order to allow for customization (e.g. launching multiple instances of the same job with different names or settings), you can add parameters. A parameter is enclosed in two curly brackets (`{{...}}`). For more details about the template syntax please refer to the [Jinja2](http://jinja.pocoo.org/) and [Jinjava](https://github.com/HubSpot/jinjava/) documentation.

**Note:** Nomad jobs have a [template stanza](https://www.nomadproject.io/docs/job-specification/template.html) that uses Consul Template. Both Jinja2 and Consul Template use curly brackets for variables by default. To resolve the conflict between template engines, one can use Nomad's [LeftDelim](https://www.nomadproject.io/api/json-jobs.html#leftdelim) and [RightDelim](https://www.nomadproject.io/api/json-jobs.html#rightdelim) parameters to change Consul Template's syntax (requires Nomad >= 0.5.5).

#### Parameter Information

The template file usually contains different parameters. These parameters must be declared in the meta information file. They can also have additional properties.

---

**UPD 1**

Document only configurable jinjava options.

**UPD 2**

Add a note about a Nomad template stanza.

Gerrrr:
fixes #224


Gerrrr:
Can you please review @FRosner ?

FRosner:
I will take a look tomorrow :)

FRosner:
Just to check if I understood correctly. We render the template using Jinjava three times:

1. When the template is loaded, we render it without any bindings to see if there is an ID parameter
1. When the template is loaded, we render it again without any bindings just to extract the parameters
1. When the template is rendered, we render with all the bindings to generate the JSON

Will the mechanism of extracting all variables still work even if they are not mentioned in the parameter infos? Will it extract only global (unbound) variables or also local ones (like the ones bound in a loop)?

server/src/main/scala/de/frosner/broccoli/templates/TemplateSource.scala
```diff
@@ -1,16 +1,73 @@
...
     */
   def loadTemplates(): Seq[Template]
+
+  def loadTemplate(templateId: String,
+                   templateString: String,
+                   templateInfo: TemplateConfig.TemplateInfo,
+                   convertDashesToUnderscores: Boolean): Try[Template] =
+    Try {
+      // find variables with dashes
+      val matcher = variablePattern.matcher(templateString)
```
FRosner: Why are we finding variables with dashes not only when we want to convert them?

server/src/main/scala/de/frosner/broccoli/templates/jinjava/JinjavaConfiguration.scala
```diff
@@ -0,0 +1,13 @@
+package de.frosner.broccoli.templates.jinjava
+
+import com.hubspot.jinjava.JinjavaConfig
+
+final case class JinjavaConfiguration(
```
FRosner: Why are we constructing so many `JinjavaConfig` objects? What is anyway happening here? You define a case class with all default parameters. Where is it injected? How can you change it?

FRosner:
I set `convert-dashes-to-underscores = true` and defined two parameters:

- `secret_parameter`
- `secret-parameter` inside an `if`

Then I didn't receive a warning, maybe because the `secret-parameter` is inside the `if`? Is that expectec behaviour? **Update**: of course it is because if it is `-` it was not inside an `if`. So if we want to use `-` legacy parameters they cannot be used in Jinja specific constructs, right? Actually this will render three parameters: `secret_parameter`, `secret`, and `parameter`.

So we should really document this conversion function well to avoid these kind of things happening unintentionally.

```
{
    "Job": {
        "Region": "global",
        "ID": "{{id}}",
        "Name": "{{id}}",
        "Type": "service",
        "Priority": 50,
        "AllAtOnce": false,
        "Datacenters": [
            "dc1"
        ],
        "Constraints": null,
        "TaskGroups": [
            {
                "Name": "http-group",
                "Count": 1,
                "Constraints": null,
                "Tasks": [
                    {
                        "Name": "http-task",
                        "Driver": "raw_exec",
                        "User": "",
                        "Config": {
                            "args": [
                                "-m",
                                "SimpleHTTPServer"
                            ],
                            "command": "/usr/bin/python"
                        },
                        "Constraints": null,
                        "Env": null,
                        "Services": [
                            {
                                "Id": "",
                                "Name": "{{id}}-web-ui-1",
                                "Tags": ["protocol-http", "{{secret_parameter}}"],
                                "PortLabel": "ui",
                                "Checks": [
                                    {
                                        "Id": "",
                                        "Name": "service: \"{{id}}-ui-1\" check",
                                        "Type": "http",
                                        "Command": "",
                                        "Args": null,
                                        "Path": "/",
                                        "Protocol": "",
                                        "Interval": 10000000000,
                                        "Timeout": 2000000000
                                    }
                                ]
                            },
                            {
                                "Id": "",
                                "Name": "{{id}}-web-ui-2",
                                "Tags": ["protocol-http"],
                                "PortLabel": "ui",
                                "Checks": [
                                    {
                                        "Id": "",
                                        "Name": "service: \"{{id}}-ui-2\" check",
                                        "Type": "http",
                                        "Command": "",
                                        "Args": null,
                                        "Path": "/doesnotexist",
                                        "Protocol": "",
                                        "Interval": 10000000000,
                                        "Timeout": 2000000000
                                    }
                                ]
                            }
                        ],
                        "Resources": {
                            "CPU": {{cpu}},
                            "MemoryMB": {% if secret-parameter %} 128 {% else %} 256 {% endif %},
                            "DiskMB": 300,
                            "IOPS": 0,
                            "Networks": [
                                {
                                    "Public": false,
                                    "CIDR": "",
                                    "ReservedPorts": [
                                        {
                                            "Label": "ui",
                                            "Value": 8000
                                        }
                                    ],
                                    "DynamicPorts": null,
                                    "IP": "",
                                    "MBits": 10
                                }
                            ]
                        },
                        "Meta": null,
                        "KillTimeout": 5000000000,
                        "LogConfig": {
                            "MaxFiles": 10,
                            "MaxFileSizeMB": 10
                        },
                        "Artifacts": null
                    }
                ],
                "RestartPolicy": {
                    "Interval": 60000000000,
                    "Attempts": 2,
                    "Delay": 15000000000,
                    "Mode": "delay"
                },
                "Meta": null
            }
        ],
        "Update": {
            "Stagger": 0,
            "MaxParallel": 0
        },
        "Periodic": null,
        "Meta": null,
        "Status": "",
        "StatusDescription": "",
        "CreateIndex": 0,
        "ModifyIndex": 0
    }
}
```

Gerrrr:
>  set convert-dashes-to-underscores = true and defined two parameters:
>
>   secret_parameter
>  secret-parameter inside an if
>
> Then I didn't receive a warning, maybe because the secret-parameter is inside the if? Is that expectec behaviour? Update: of course it is because if it is - it was not inside an if. So if we want to use - legacy parameters they cannot be used in Jinja specific constructs, right? Actually this will render three parameters: secret_parameter, secret, and parameter.

We have `convert-dashes-to-underscores` option to convert dashes that were before considered to be a part of the variable name to dashes because Jinja2 would think that those dashes are actually a subtraction operations. Before this PR the only allowed variable format was like `{{id}}`, so opening curly brackets, a variable name and closing curly brackets (**NB** that there are no spaces).

Broccoli raises warnings (or crashes, depending on the `convert-dashes-to-underscores` flag) if it finds a variable in the legacy format. In the job file you posted the variables `secret` and `parameter` are inside the `if` statement that was not supported before, so there is no reason to raise a warning. Following the same logic, there should be no warning for variable usage like `{{ secret-variable }}`.

FRosner:
>>  set convert-dashes-to-underscores = true and defined two parameters:
>>
>>   secret_parameter
>>  secret-parameter inside an if
>>
>> Then I didn't receive a warning, maybe because the secret-parameter is inside the if? Is that expectec behaviour? Update: of course it is because if it is - it was not inside an if. So if we want to use - legacy parameters they cannot be used in Jinja specific constructs, right? Actually this will render three parameters: secret_parameter, secret, and parameter.
>
> We have `convert-dashes-to-underscores` option to convert dashes that were before considered to be a part of the variable name to dashes because Jinja2 would think that those dashes are actually a subtraction operations. Before this PR the only allowed variable format was like `{{id}}`, so opening curly brackets, a variable name and closing curly brackets (**NB** that there are no spaces).
>
>Broccoli raises warnings (or crashes, depending on the `convert-dashes-to-underscores` flag) if it finds a variable in the legacy format. In the job file you posted the variables `secret` and `parameter` are inside the `if` statement that was not supported before, so there is no reason to raise a warning. Following the same logic, there should be no warning for variable usage like `{{ secret-variable }}`.

That's what I figured. Makes sense. Thank you!

Gerrrr:
>Just to check if I understood correctly. We render the template using Jinjava three times:
>
>   When the template is loaded, we render it without any bindings to see if there is an ID parameter
>    When the template is loaded, we render it again without any bindings just to extract the parameters
>    When the template is rendered, we render with all the bindings to generate the JSON

Right. First, we load it to find variables in the legacy format that can contain dashes. Then, we load it again to check if the `id` parameter is there. Then, we actually render the template with the instance information. So the first two parsings happen only once per template while the rendering happens for each template instance.

>Will the mechanism of extracting all variables still work even if they are not mentioned in the parameter infos? Will it extract only global (unbound) variables or also local ones (like the ones bound in a loop)?

Quick test:
```scala
      val template = "{{id}} {{ global_var }} {% for x in [1 2 3] %}{{x}}{% endfor %}"
      val jinjava = new Jinjava()
      val renderResult = jinjava.renderForResult(template, Map("id" -> "abc", "global_var" -> "456"))
      println(renderResult.getOutput)
      println(renderResult.getContext.getResolvedValues)
```

prints:

```
abc 456 123
[x, id, global_var]
```

So it returns both global and local variables. Should I introduce explicit variable annotation in the `template.conf` also in this PR?

FRosner:
>So it returns both global and local variables. Should I introduce explicit variable annotation in the `template.conf` also in this PR?

Does that mean that the UI will also show you an input field for `x`? What happens if you provide something there?

Gerrrr:
Yes, you are right. Nice catch! I am going to fix it.

FRosner (changes requested):
> Yes, you are right. Nice catch! I am going to fix it.
Please do that. Please also document it in the ticket that this needs to be documented.

Gerrrr:
> So it returns both global and local variables. Should I introduce explicit variable annotation in the template.conf also in this PR?
> 
> Does that mean that the UI will also show you an input field for x? What happens if you provide something there?

Since I am going to introduce explicit variable declaration in the `template.conf`, do we really need `convert-dashes-to-underscores` flag? If all the variables are to be reviewed anyway, they can be as well renamed. Broccoli can still validate while loading templates that the declared variables do not contain dashes.

FRosner:
> Since I am going to introduce explicit variable declaration in the template.conf, do we really need convert-dashes-to-underscores flag? If all the variables are to be reviewed anyway, they can be as well renamed. Broccoli can still validate while loading templates that the declared variables do not contain dashes.

Yes we do because the problem is not about reviewing and renaming taking effort, it is about the problem of template migration. Renaming it will make it a new variable and the UI has no support for upgrading like this. It will be a nightmare.

> So it returns both global and local variables. Should I introduce explicit variable annotation in the template.conf also in this PR?

I think we should because if we merge this PR otherwise we introduced a bug (the UI showing `x` as an input field although it is a locally bound variable).

FRosner:
How are you going to "introduce" explicit variable declaration? There is not much to introduce. You just have to get the variables that exist from there and we need to document that all variables need to exist there. Or am I missing something?

server/src/main/scala/de/frosner/broccoli/templates/TemplateRenderer.scala
```diff
@@ -26,22 +31,30 @@ class TemplateRenderer(defaultType: ParameterType) {
     sanitized
   }
 
-  def renderJson(instance: Instance): JsValue = {
+  def renderForResult(instance: Instance): RenderResult = {
```
FRosner: Can we make some of the methods private?

server/src/main/scala/de/frosner/broccoli/templates/TemplateRenderer.scala
```diff
@@ -26,22 +31,30 @@ class TemplateRenderer(defaultType: ParameterType) {
...
-      case (intermediateTemplate, (parameter, value)) =>
-        intermediateTemplate.replaceAllLiterally(s"{{$parameter}}", sanitize(parameter, value, parameterInfos))
+    jinjava.renderForResult(template.template, parameterValues)
+  }
+
+  def renderJson(instance: Instance): JsValue = {
+    val renderResult = renderForResult(instance)
+    val fatalErrors = renderResult.getErrors.filter(error => error.getSeverity == ErrorType.FATAL)
+
+    if (!fatalErrors.isEmpty()) {
```
FRosner: should be `fatalErrors.nonEmpty` instead, without braces

FRosner:
When starting Broccoli with this changes, instances which were previously created can't be loaded:

```
play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[IllegalArgumentException: requirement failed: The given parameters values (Set(id, sleep-count, time)) need to match the ones in the template (Set(time, id, enabled)).]]
	at play.api.http.HttpErrorHandlerExceptions$.throwableToUsefulException(HttpErrorHandler.scala:293)
	at play.api.http.DefaultHttpErrorHandler.onServerError(HttpErrorHandler.scala:220)
	at play.api.GlobalSettings$class.onError(GlobalSettings.scala:160)
	at play.api.DefaultGlobal$.onError(GlobalSettings.scala:188)
	at play.api.http.GlobalSettingsHttpErrorHandler.onServerError(HttpErrorHandler.scala:100)
	at play.core.server.netty.PlayRequestHandler$$anonfun$handle$1.applyOrElse(PlayRequestHandler.scala:129)
	at play.core.server.netty.PlayRequestHandler$$anonfun$handle$1.applyOrElse(PlayRequestHandler.scala:127)
	at scala.concurrent.Future$$anonfun$recoverWith$1.apply(Future.scala:346)
	at scala.concurrent.Future$$anonfun$recoverWith$1.apply(Future.scala:345)
	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:36)
	at play.api.libs.iteratee.Execution$trampoline$.executeScheduled(Execution.scala:109)
	at play.api.libs.iteratee.Execution$trampoline$.execute(Execution.scala:71)
	at scala.concurrent.impl.CallbackRunnable.executeWithValue(Promise.scala:44)
	at scala.concurrent.impl.Promise$DefaultPromise.tryComplete(Promise.scala:252)
	at scala.concurrent.Promise$class.complete(Promise.scala:55)
	at scala.concurrent.impl.Promise$DefaultPromise.complete(Promise.scala:157)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.run(Future.scala:23)
	at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:39)
	at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(AbstractDispatcher.scala:415)
	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
	at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
	at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
Caused by: java.lang.IllegalArgumentException: requirement failed: The given parameters values (Set(id, sleep-count, time)) need to match the ones in the template (Set(time, id, enabled)).
	at scala.Predef$.require(Predef.scala:224)
	at de.frosner.broccoli.models.Instance.requireParameterValueConsistency(Instance.scala:13)
	at de.frosner.broccoli.models.Instance.<init>(Instance.scala:20)
	at de.frosner.broccoli.models.Instance$$anonfun$4.apply(Instance.scala:36)
	at de.frosner.broccoli.models.Instance$$anonfun$4.apply(Instance.scala:36)
	at play.api.libs.functional.FunctionalBuilder$CanBuild3$$anonfun$apply$7.apply(Products.scala:77)
	at play.api.libs.functional.FunctionalBuilder$CanBuild3$$anonfun$apply$7.apply(Products.scala:77)
	at play.api.libs.json.JsResult$class.map(JsResult.scala:77)
	at play.api.libs.json.JsSuccess.map(JsResult.scala:9)
	at play.api.libs.json.Reads$$anonfun$map$1.apply(Reads.scala:46)
	at play.api.libs.json.Reads$$anonfun$map$1.apply(Reads.scala:46)
	at play.api.libs.json.Reads$$anon$8.reads(Reads.scala:126)
	at play.api.libs.json.JsValue$class.validate(JsValue.scala:18)
	at play.api.libs.json.JsObject.validate(JsValue.scala:76)
	at play.api.libs.json.JsReadable$class.as(JsReadable.scala:21)
	at play.api.libs.json.JsObject.as(JsValue.scala:76)
	at de.frosner.broccoli.instances.storage.filesystem.FileSystemInstanceStorage$$anonfun$6.apply(FileSystemInstanceStorage.scala:44)
	at de.frosner.broccoli.instances.storage.filesystem.FileSystemInstanceStorage$$anonfun$6.apply(FileSystemInstanceStorage.scala:44)
	at scala.util.Success$$anonfun$map$1.apply(Try.scala:237)
	at scala.util.Try$.apply(Try.scala:192)
	at scala.util.Success.map(Try.scala:237)
	at de.frosner.broccoli.instances.storage.filesystem.FileSystemInstanceStorage.readInstanceImpl(FileSystemInstanceStorage.scala:44)
	at de.frosner.broccoli.instances.storage.filesystem.FileSystemInstanceStorage$$anonfun$readInstancesImpl$2$$anonfun$apply$3.apply(FileSystemInstanceStorage.scala:74)
	at de.frosner.broccoli.instances.storage.filesystem.FileSystemInstanceStorage$$anonfun$readInstancesImpl$2$$anonfun$apply$3.apply(FileSystemInstanceStorage.scala:73)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
	at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:186)
	at scala.collection.TraversableLike$class.map(TraversableLike.scala:234)
	at scala.collection.mutable.ArrayOps$ofRef.map(ArrayOps.scala:186)
	at de.frosner.broccoli.instances.storage.filesystem.FileSystemInstanceStorage$$anonfun$readInstancesImpl$2.apply(FileSystemInstanceStorage.scala:73)
	at de.frosner.broccoli.instances.storage.filesystem.FileSystemInstanceStorage$$anonfun$readInstancesImpl$2.apply(FileSystemInstanceStorage.scala:79)
	at scala.util.Success$$anonfun$map$1.apply(Try.scala:237)
	at scala.util.Try$.apply(Try.scala:192)
	at scala.util.Success.map(Try.scala:237)
	at de.frosner.broccoli.instances.storage.filesystem.FileSystemInstanceStorage.readInstancesImpl(FileSystemInstanceStorage.scala:73)
	at de.frosner.broccoli.instances.storage.InstanceStorage$$anonfun$readInstances$3$$anonfun$apply$3.apply(InstanceStorage.scala:50)
	at de.frosner.broccoli.instances.storage.InstanceStorage$$anonfun$readInstances$3$$anonfun$apply$3.apply(InstanceStorage.scala:50)
	at de.frosner.broccoli.instances.storage.InstanceStorage$class.de$frosner$broccoli$instances$storage$InstanceStorage$$ifNotClosed(InstanceStorage.scala:15)
	at de.frosner.broccoli.instances.storage.InstanceStorage$$anonfun$readInstances$3.apply(InstanceStorage.scala:49)
	at de.frosner.broccoli.instances.storage.InstanceStorage$$anonfun$readInstances$3.apply(InstanceStorage.scala:49)
	at de.frosner.broccoli.logging$.logExecutionTime(logging.scala:23)
	at de.frosner.broccoli.instances.storage.InstanceStorage$class.readInstances(InstanceStorage.scala:52)
	at de.frosner.broccoli.instances.storage.filesystem.FileSystemInstanceStorage.readInstances(FileSystemInstanceStorage.scala:13)
	at de.frosner.broccoli.instances.storage.filesystem.FileSystemInstanceStorage.readInstancesImpl(FileSystemInstanceStorage.scala:59)
	at de.frosner.broccoli.instances.storage.InstanceStorage$$anonfun$readInstances$1$$anonfun$apply$2.apply(InstanceStorage.scala:38)
	at de.frosner.broccoli.instances.storage.InstanceStorage$$anonfun$readInstances$1$$anonfun$apply$2.apply(InstanceStorage.scala:38)
	at de.frosner.broccoli.instances.storage.InstanceStorage$class.de$frosner$broccoli$instances$storage$InstanceStorage$$ifNotClosed(InstanceStorage.scala:15)
	at de.frosner.broccoli.instances.storage.InstanceStorage$$anonfun$readInstances$1.apply(InstanceStorage.scala:37)
	at de.frosner.broccoli.instances.storage.InstanceStorage$$anonfun$readInstances$1.apply(InstanceStorage.scala:37)
	at de.frosner.broccoli.logging$.logExecutionTime(logging.scala:23)
	at de.frosner.broccoli.instances.storage.InstanceStorage$class.readInstances(InstanceStorage.scala:40)
	at de.frosner.broccoli.instances.storage.filesystem.FileSystemInstanceStorage.readInstances(FileSystemInstanceStorage.scala:13)
	at de.frosner.broccoli.services.InstanceService.initializeInstancesMap(InstanceService.scala:152)
	at de.frosner.broccoli.services.InstanceService.de$frosner$broccoli$services$InstanceService$$instances(InstanceService.scala:171)
	at de.frosner.broccoli.services.InstanceService.getInstances(InstanceService.scala:232)
	at de.frosner.broccoli.controllers.InstanceController$$anonfun$list$3.apply(InstanceController.scala:162)
	at de.frosner.broccoli.controllers.InstanceController$$anonfun$list$3.apply(InstanceController.scala:162)
	at scala.Option.getOrElse(Option.scala:121)
	at de.frosner.broccoli.controllers.InstanceController$.list(InstanceController.scala:162)
	at de.frosner.broccoli.controllers.WebSocketController$$anonfun$requestToSocket$1.apply(WebSocketController.scala:79)
	at de.frosner.broccoli.controllers.WebSocketController$$anonfun$requestToSocket$1.apply(WebSocketController.scala:34)
	at jp.t2v.lab.play2.auth.BroccoliWebsocketSecurity$class.withSecurity(BroccoliWebsocketSecurity.scala:50)
	at de.frosner.broccoli.controllers.WebSocketController.withSecurity(WebSocketController.scala:20)
	at de.frosner.broccoli.controllers.WebSocketController.requestToSocket(WebSocketController.scala:34)
	at de.frosner.broccoli.controllers.WebSocketController$$anonfun$socket$1.apply(WebSocketController.scala:83)
	at de.frosner.broccoli.controllers.WebSocketController$$anonfun$socket$1.apply(WebSocketController.scala:83)
	at scala.Function1$$anonfun$andThen$1.apply(Function1.scala:52)
	at play.api.mvc.WebSocket$$anonfun$acceptOrResult$1.apply(WebSocket.scala:250)
	at play.api.mvc.WebSocket$$anonfun$acceptOrResult$1.apply(WebSocket.scala:249)
	at play.api.mvc.WebSocket$$anon$1.apply(WebSocket.scala:43)
	at play.core.routing.TaggingInvoker$$anonfun$call$1.apply(HandlerInvoker.scala:48)
	at play.core.routing.TaggingInvoker$$anonfun$call$1.apply(HandlerInvoker.scala:48)
	at play.api.mvc.WebSocket$$anon$1.apply(WebSocket.scala:43)
	at play.core.server.netty.PlayRequestHandler$$anonfun$3.apply(PlayRequestHandler.scala:113)
	at play.core.server.netty.PlayRequestHandler$$anonfun$3.apply(PlayRequestHandler.scala:113)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.liftedTree1$1(Future.scala:24)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.run(Future.scala:24)
	... 6 common frames omitted
```

FRosner:
Is

```
    jinjava {
      fail-on-unknown-tokens = true
    }
```

working as expected?

Gerrrr:
Before merging this PR we need to document the breaking changes:

* `template.conf` must include all variables in the parameter section.
* template variables cannot contain dashes any longer because we render the templates with `jinja2` and it will treat dashes as a subtraction operation. If you still have dashes in the variable names, you need to set `broccoli.templates.convert-dashes-to-underscores` to `true`.

Further TODOs:

* Check that the examples in `README` and wiki still work, otherwise adapt them.
* Document new `broccoli.templates.jinjava` properties.

FRosner:
Regarding the doc change, please copy the corresponding wiki section into the ticket (#224), change it the way it needs to be so we can include it later.

FRosner:
I found the following problem, which users will encounter when upgrading:

1. Add new template: [sleep.zip](https://github.com/FRosner/cluster-broccoli/files/1751564/sleep.zip)
2. Add new instances: [instances.zip](https://github.com/FRosner/cluster-broccoli/files/1751569/instances.zip)
3. Enable dash-conversion
3. Start Broccoli in the new version (this PR)
4. ```13:00:54.974 [ERROR] [pool-13-thread-1] [d.f.b.services.InstanceService] - Failed to load the instances: java.lang.IllegalArgumentException: requirement failed: The given parameters values (Set(id, sleep-count, time)) need to match the ones in the template (Set(time, id, enabled)).```

I think there are three problems hidden in the upgrade process:

1. When upgrading you need to explicitly specify all possible parameter values in the `template.conf` => Solved by documentation
2. When converting dashes you need to also convert it in the `template.conf` object (`TemplateInfo`), because otherwise you are inconsistent. => You have to change your code, I guess
3. A persisted instance stores its template. If this template did not have all parameters in `template.conf`, the instance will not even succeed to be loaded. => I see two options here: Either we are more relaxed when loading "invalid" instances in order to allow the user to upgrade to the latest template, or we provide an upgrade script for the admins that will go through the instance storage and "fix" the templates by at least adding the stub for each missing parameters to the parameter infos. I prefer option 2, as otherwise everybody will be forced to upgrade to the latest template version, which might also contain other changes they don't want.

Gerrrr:
>Is
>
>    jinjava {
>      fail-on-unknown-tokens = true
>    }
>
> working as expected?

It is the option to set a jinjava's `failOnUnknownTokens` flag. From jinjava's [changelog](https://github.com/HubSpot/jinjava/blob/ff924a8752b741116c03505a02e6568276019c37/CHANGES.md#L98): 
> added failOnUnknownTokens mode which is similar to Jinja's StrictUndefined

Here is the behavior of Jinja's `StrictUndefined` ([source](http://jinja.pocoo.org/docs/2.10/api/)):
> StrictUndefined which disallows all operations beside testing if it’s an undefined object.

AFAIU it throws an exception while rendering if a tag or a variable is not defined in the scope.

Gerrrr:
Documentation changes:

https://github.com/FRosner/cluster-broccoli/wiki/Upgrading-to-0.8.0

**Templates**

Templates are now rendered with [Jinjava template engine](https://github.com/HubSpot/jinjava). Because of this, there are several changes breaking backwards compatibility:
* Variables can no longer contain dashes. In order to automatically convert dashes in the variable names in instances and templates, you can set `broccoli.templates.convert-dashes-to-underscores` to `true`.
* Broccoli no longer automatically discovers template variables. You have to explicitly specify all variables in `template.conf`. To fix existing instances, please run the upgrade script:
    - `nomad stop cluster-broccoli`
    - `python script/instances-0.7.0-to-0.8.0.py /path/to/broccoli/instances`
    - `nomad start cluster-broccoli`

**Configuration Changes**

It is possible to set [Jinjava config settings](https://static.javadoc.io/com.hubspot.jinjava/jinjava/2.4.0/index.html?com/hubspot/jinjava/JinjavaConfig.html) through Broccoli. The default values are kept the same as in Jinjava, unless stated otherwise. For documentation on those parameters please refer to the original project https://github.com/HubSpot/jinjava.

Here is the list of Jinjava settings configurable through Broccoli:

* `broccoli.templates.jinjava.trim-blocks` (defaults to `false`)
* `broccoli.templates.jinjava.lstrip-blocks` (defaults to `false`)
* `broccoli.templates.jinjava.max-render-depth` (defaults to `10`)
* `broccoli.templates.jinjava.read-only-resolver` (defaults to `true`)
* `broccoli.templates.jinjava.enable-recursive-macro-calls` (defaults to `false`)
* `broccoli.templates.jinjava.fail-on-unknown-tokens` (defaults to `true`)
* `broccoli.templates.jinjava.max-output-size` (defaults to `0`)
* `broccoli.templates.jinjava.nested-interpretation-enabled` (defaults to `true`)

---

https://github.com/FRosner/cluster-broccoli/wiki/Templates

### Nomad JSON Template File

The template file is a Jinja2 template of a [Nomad job file in JSON format](https://www.nomadproject.io/docs/jobspec/json.html). In order to allow for customization (e.g. launching multiple instances of the same job with different names or settings), you can add parameters. A parameter is enclosed in two curly brackets (`{{...}}`). For more details about the template syntax please refer to the [Jinja2](http://jinja.pocoo.org/) and [Jinjava](https://github.com/HubSpot/jinjava/) documentation.

**Note:** Nomad jobs have a [template stanza](https://www.nomadproject.io/docs/job-specification/template.html) that uses Consul Template. Both Jinja2 and Consul Template use curly brackets for variables by default. To resolve the conflict between template engines, one can use Nomad's [left_delimiter](https://www.nomadproject.io/docs/job-specification/template.html#left_delimiter) and [right_delimiter](https://www.nomadproject.io/docs/job-specification/template.html#right_delimiter) parameters to change Consul Template's syntax.

#### Parameter Information

The template file usually contains different parameters. These parameters must be declared in the meta information file. They can also have additional properties.

---

**UPD 1**

Document only configurable jinjava options.

**UPD 2**

Add a note about a Nomad template stanza.

FRosner:
I can't seem to find the config values you put here in the `reference.conf` file. Should we put them there? IMHO the reference configuration should contain all defaults.

Gerrrr:
You are right, it is actually a good idea. I'll add the jinjava config values to `reference.conf`.

server/src/main/resources/reference.conf
```diff
@@ -47,7 +47,14 @@ broccoli {
     convert-dashes-to-underscores = false
     # jinjava configuration
```
FRosner: Sorry to bug again. Can we put a bit more documentation? I understand that we don't want to duplicate the Jinjava documentation here but at least link to it or make it more obvious where to look for the explanation of those values?

FRosner: Also write what this part is used for. If you see the Jinjava block, it's not obvious if you need to configure it or so. In HOCON you can make block comments as well if you feel like it I think.

server/src/main/scala/de/frosner/broccoli/services/InstanceService.scala
```diff
@@ -199,6 +199,36 @@ class InstanceService @Inject()(nomadClient: NomadClient,
...
+    Try {
+      require(newParameterValues("id") == instance.parameterValues("id"),
+              s"The parameter value 'id' must not be changed.")
+
+      val newInstance = Instance(
+        id = instance.id,
+        template = newTemplate,
+        parameterValues = newParameterValues
+      )
+      require(!templateRenderer.renderForResult(newInstance).hasErrors)
```
sohaibiftikhar, line 228: @Gerrrr Should we have more meaningful errors here? If this condition fails the frontend shows "requirement failed" which does not look very helpful.

sohaibiftikhar (commented):
I fixed partially in a [commit](https://github.com/FRosner/cluster-broccoli/pull/311/commits/0ccab551302a6f62392157e4268ac2565b59a065).

server/src/main/scala/de/frosner/broccoli/services/InstanceService.scala
FRosner: Yes please add a meaningful error message.

FRosner (changes requested):
LGTM. As discussed we should add proper validation of parameter names in the `template.conf` during template parsing.

FRosner (approved):

PR #317, opened 2017-11-03
Add parameter order-index

### LINKED ISSUE #312: Support order index for parameters
## Problem

Parameter order right now is determined by the parameter ID. This is bad because reordering parameters requires changing the ID, which basically means removing the parameter and adding a new one. This renders template upgrades really difficult.

## Solution

Introduce a new attribute for each parameter that indicates the position in the ordering. It should be an integer and the parameters will be sorted by this index and then by ID if there is a tie. If a parameter does not have an index it will be shown in the end.

### Example

```
parameters = {
  "username" = {
    name = "Username"
    type = string
    order-index = 1
  }
  "password" = {
    name = "Password"
    secret = true
    order-index = 2
  }
  "http-port" = {
    name = "HTTP Port"
    default = 80
    order-index = 3
  }
}
```

FRosner:
fixes #312

### DIFF
--- a/http-api-tests/broccoli-nomad-consul/api-v1-instances-service-status/expected/response-data
+++ b/http-api-tests/broccoli-nomad-consul/api-v1-instances-service-status/expected/response-data
@@ -27,7 +27,7 @@
         "type": "string"
       }
     },
-    "version": "e1ee8d8dd92a3b1bed4392da1cdea8c6"
+    "version": "1721c964c355a8590c1c994d443ca1cf"
   },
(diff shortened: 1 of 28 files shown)


FRosner:
I am going to reduce code duplication a bit

FRosner:
Much better. Feel free to review now @Gerrrr  ))

Gerrrr:
Sure, I am going to take a look during this weekend.

Gerrrr (approved):
LGTM!

PR #339, opened 2017-12-06
Issue 337

### LINKED ISSUE #337: Use atomic move operation in file system instance storage
### Problem

Right now, if Broccoli crashes during a [write operation](https://github.com/FRosner/cluster-broccoli/blob/1165d42c61c9bb47c52a2ee19c21a2e8ae09c4ed/server/src/main/scala/de/frosner/broccoli/instances/storage/filesystem/FileSystemInstanceStorage.scala#L80) using the file system instance storage, this might lead to corrupted data.

### Solution

- Create `<instance-id>.json_tmp` inside the instance dir and write the content there.
- Use [`Files.move`](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#move(java.nio.file.Path,%20java.nio.file.Path,%20java.nio.file.CopyOption...)) with [`StandardCopyOption.ATOMIC_MOVE`](https://docs.oracle.com/javase/7/docs/api/java/nio/file/StandardCopyOption.html#ATOMIC_MOVE) to overwrite `<instance-id>.json` afterwards.

IshmeetKaur:
Use atomic move operation in file system instance storage

fixes #337

### DIFF
--- a/server/src/main/scala/de/frosner/broccoli/instances/storage/filesystem/FileSystemInstanceStorage.scala
+++ b/server/src/main/scala/de/frosner/broccoli/instances/storage/filesystem/FileSystemInstanceStorage.scala
@@ -1,6 +1,7 @@
 package de.frosner.broccoli.instances.storage.filesystem
 
 import java.io._
+import java.nio.file._
 
 import de.frosner.broccoli.instances.storage.InstanceStorage
 import de.frosner.broccoli.models.Instance
@@ -25,6 +26,8 @@ class FileSystemInstanceStorage(storageDirectory: File) extends InstanceStorage 
   if (!lock.createNewFile()) {
     throw new IllegalStateException(s"Cannot lock $storageDirectory. Is there another Broccoli instance running?")
   }
+
+  def idToFileTemp(id: String): File = new File(storageDirectory, id + ".json_tmp")
 
   def idToFile(id: String): File = new File(storageDirectory, id + ".json")
 
@@ -79,11 +82,12 @@ class FileSystemInstanceStorage(storageDirectory: File) extends InstanceStorage 
   @volatile
   override def writeInstanceImpl(instance: Instance): Try[Instance] = {
     val id = instance.id
-    val file = idToFile(id)
+    val file = idToFileTemp(id)
     val printStream = Try(new PrintStream(new FileOutputStream(file)))
     val afterWrite = printStream.map(_.append(Json.toJson(instance).toString()))
-    printStream.map(_.close())
-    afterWrite.map(_ => instance)
+    val finishWrite = printStream.map(_.close())
+    val finishMove = finishWrite.flatMap(_ => Try(Files.move(idToFileTemp(id).toPath, idToFile(id).toPath, StandardCopyOption.ATOMIC_MOVE)))
+    finishMove.map(_ => instance)
   }
 
   @volatile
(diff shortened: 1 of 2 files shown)


server/src/main/scala/de/frosner/broccoli/instances/storage/filesystem/FileSystemInstanceStorage.scala
```diff
@@ -79,11 +82,13 @@ class FileSystemInstanceStorage(storageDirectory: File) extends InstanceStorage
   @volatile
   override def writeInstanceImpl(instance: Instance): Try[Instance] = {
     val id = instance.id
-    val file = idToFile(id)
```
FRosner, line 82: Please call it `tempFile` to make it more obvious that it is temporary.

server/src/main/scala/de/frosner/broccoli/instances/storage/filesystem/FileSystemInstanceStorage.scala
```diff
@@ -79,11 +82,13 @@ class FileSystemInstanceStorage(storageDirectory: File) extends InstanceStorage
...
+    val file = idToFileTemp(id)
     val printStream = Try(new PrintStream(new FileOutputStream(file)))
     val afterWrite = printStream.map(_.append(Json.toJson(instance).toString()))
-    printStream.map(_.close())
-    afterWrite.map(_ => instance)
+    val finishWrite = printStream.map(_.close())
```
FRosner: I guess we should map based on `afterWrite` to also catch if appending breaks.

server/src/main/scala/de/frosner/broccoli/instances/storage/filesystem/FileSystemInstanceStorage.scala
```diff
@@ -79,11 +82,13 @@ class FileSystemInstanceStorage(storageDirectory: File) extends InstanceStorage
...
+    val finishMove = finishWrite.flatMap(_ =>
+      Try(Files.move(idToFileTemp(id).toPath, idToFile(id).toPath, StandardCopyOption.ATOMIC_MOVE)))
```
FRosner: Please reuse the variable you created instead of again generating the file name with `idToFileTemp(id)`.

server/src/main/scala/de/frosner/broccoli/instances/storage/filesystem/FileSystemInstanceStorage.scala
```diff
@@ -26,6 +27,8 @@ class FileSystemInstanceStorage(storageDirectory: File) extends InstanceStorage
     throw new IllegalStateException(s"Cannot lock $storageDirectory. Is there another Broccoli instance running?")
   }
 
+  def idToFileTemp(id: String): File = new File(storageDirectory, id + ".json_tmp")
```
FRosner: Should be called `idToTempFile`, no?

FRosner (changes requested):

server/src/main/scala/de/frosner/broccoli/instances/storage/filesystem/FileSystemInstanceStorage.scala
FRosner: Sorry I might have been confused here. If we map it on `finishWrite` we'll not have the error when appending.

Let's quickly look at it together on Monday.

IshmeetKaur: Yes.. Sure

FRosner (dismissed):

FRosner (approved):

PR #344, opened 2018-02-13
#342 periodic job allocation info

### LINKED ISSUE #342: Periodic job allocation information
## Problem

When starting a periodic job, the individual runs are also nomad jobs. However they have no allocation information displayed so you cannot really see the resource consumption or logs.

## Solution

Add the same allocation information that exist for instances also for periodic runs.

## Documentation Change

- [x] We have to adjust the HTTP API serving logs to also support periodic task logs

## ToDo

- [x] Check the following => Done. Problem was that if getting the nomad job for the periodic job fails, it would never even attempt to get the periodic jobs.
  - Start periodic job, wait for one or two allocations to be placed
  - Stop the job => dead allocations are gone
  - Start the job => dead allocations reappear
- [x] #190
- [x] Allocation info
- [x] Log download

## Links

- relates to #340 
- is annoying without #190 fixed

FRosner:
fixes #342

### DIFF
--- a/server/src/main/scala/de/frosner/broccoli/controllers/InstanceController.scala
+++ b/server/src/main/scala/de/frosner/broccoli/controllers/InstanceController.scala
@@ -68,8 +68,39 @@ case class InstanceController @Inject()(
   ): Action[Unit] =
     AsyncStack(parse.empty) { implicit request =>
       instances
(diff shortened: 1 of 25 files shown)


FRosner:
Gonna rebase, restructure and reopen a new PR

PR #351, opened 2018-02-19
#350 websocket not being refreshed

### LINKED ISSUE #350: Websocket sometimes not refreshed
## Problem

- Start Broccoli
- Be unlucky
- Websocket not being refreshed for clients

## Root Cause

In the code we initialize

```scala
private var connections: Map[String, (Account, Set[Concurrent.Channel[Msg]])] = Map.empty
```

after we use it in another thread:

```scala
private val scheduler = new ScheduledThreadPoolExecutor(1)
private val task = new Runnable {
  def run() = {
    val numChannels = connections.foldLeft(0) {
      case (n, (id, (account, channels))) => n + channels.size
    }
    logging.logExecutionTime(s"Updating $numChannels websocket channels for ${connections.size} sessions") {
      broadcast { user =>
        Json.toJson(OutgoingMessage.AboutInfoMsg(AboutController.about(aboutInfoService, user)))
      }
      broadcast { _ =>
        Json.toJson(OutgoingMessage.ListTemplates(TemplateController.list(templateService)))
      }
      broadcast { user =>
        Json.toJson(OutgoingMessage.ListInstances(InstanceController.list(None, user, instanceService)))
      }
    }(log.info(_))
  }
}
private val scheduledTask = scheduler.scheduleWithFixedDelay(task, 0, 1, TimeUnit.SECONDS)
```

I didn't notice it before as the scheduling usually takes enough time but you might be unlucky.

## Fix

Move the initialization before the usage.

FRosner:
fixes #350

### DIFF
--- a/server/src/main/scala/de/frosner/broccoli/services/WebSocketService.scala
+++ b/server/src/main/scala/de/frosner/broccoli/services/WebSocketService.scala
@@ -24,6 +24,9 @@ class WebSocketService @Inject()(templateService: TemplateService,
 
   log.info(s"Starting $this")
 
+  @volatile
+  private var connections: Map[String, (Account, Set[Concurrent.Channel[Msg]])] = Map.empty
+
   private val scheduler = new ScheduledThreadPoolExecutor(1)
   private val task = new Runnable {
     def run() = {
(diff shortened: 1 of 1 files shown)


aurelianrl (approved):
LGTM

PR #354, opened 2018-03-14
Stop Button for completed periodic jobs shows a trash can

### LINKED ISSUE #355: Trash can button for stopped and dead periodic jobs

Cubic-BB8:
fixes #355

### DIFF
--- a/webui/src/Views/PeriodicRunsView.elm
+++ b/webui/src/Views/PeriodicRunsView.elm
@@ -104,16 +104,22 @@ row instanceId instanceTasks periodicRun =
                                 ]
                                 [ JobStatusView.view "" periodicRun.status
                                 , text " "
-                                , (if (periodicRun.status == JobStatus.JobRunning) then
-                                    iconButton
-                                        "btn btn-default btn-xs"
-                                        "glyphicon glyphicon-stop"
-                                        "Stop and Delete Instance"
-                                   else
+                                , (if
+                                    (periodicRun.status
+                                        == JobStatus.JobStopped
+                                        || periodicRun.status
+                                        == JobStatus.JobDead
+                                    )
+                                   then
                                     iconButton
                                         "btn btn-default btn-xs"
                                         "glyphicon glyphicon-trash"
                                         "Delete Instance"
+                                   else
+                                    iconButton
+                                        "btn btn-default btn-xs"
+                                        "glyphicon glyphicon-stop"
+                                        "Stop and Delete Instance"
                                   )
                                     (List.append
                                         [ onClick (StopPeriodicJobs instanceId [ periodicRun.jobName ])
--- a/webui/src/Views/PeriodicRunsView.elm
+++ b/webui/src/Views/PeriodicRunsView.elm
@@ -104,10 +104,19 @@ row instanceId instanceTasks periodicRun =
                                 ]
                                 [ JobStatusView.view "" periodicRun.status
                                 , text " "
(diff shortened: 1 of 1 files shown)


webui/src/Views/PeriodicRunsView.elm
```diff
@@ -104,10 +104,19 @@ row instanceId instanceTasks periodicRun =
...
+                                        iconButton
+                                            "btn btn-default btn-xs"
+                                            "glyphicon glyphicon-stop"
+                                            "Stop and Delete Instance"
+                                    else
+                                        iconButton
+                                            "btn btn-default btn-xs"
+                                            "glyphicon glyphicon-trash"
+                                            "Delete Instance"
+                                    )
```
FRosner: Does it make sense to show a trash can as default? IMO when you want to stop a pending job, a stop button makes sense. I would show the trash can only for jobs that are dead (so done basically), and keep the stop button default.

What do you think?

FRosner (changes requested):

webui/src/Views/PeriodicRunsView.elm
Cubic-BB8: @FRosner 
IMHO:
you trash the instance in any case, no matter what is its current status. So, trash should be shown to user, otherwise this "STOP" with side effect as really DELETE is misleading even more. 

Nonetheless, check this [fix](https://github.com/FRosner/cluster-broccoli/pull/354/commits/f1b634cfc8964e3dc0034f22745b0da3d7bfb66a)

FRosner: Thanks for the fix. Will a running periodic job that you stop also be completely removed or marked as dead and ready for garbage collection? If a periodic job which you invoke stop on is removed independently of its current status, we can indeed consider showing the trash can. But then we should show it always, even for running jobs.

Cubic-BB8: @FRosner ,
> Will a running periodic job that you stop also be completely removed or marked as dead and ready for garbage collection

With current version of Nomad it will be trashed. There is no option to just STOP it.

> But then we should show it always, even for running jobs

This was my initial point, but you should better align with other opinions.

In current state code satisfies requirements provided in the ticket.

FRosner: Ok. Given what you say I suggest we replace the stop button with a trash button completely, removing the condition on the job status. The title could say "Stop and delete periodic run" or something like this.

You can align with Nico but to me this sounds like the best option.

Sorry for going back-and-forth but I found the discussion super useful! 👍

FRosner: Having said that I am also fine with the current implementation and happy to merge it like this. You tell me how you guys decided.

FRosner (approved):

webui/src/Views/PeriodicRunsView.elm
Cubic-BB8: @FRosner 
Regarding 
> You can align with Nico but to me this sounds like the best option.
> Sorry for going back-and-forth but
 
Btw, Nico is not in the office. There are requirements in the ticket, which was actually refined as well. The code satisfies this requirements and from my point of view there is no reason not to take it for next steps.

In case this finding will be reviewed - it should end up as a new ticket. In any case it is not to much work.

FRosner: Works for me!

PR #357, opened 2018-03-16
[WIP] Add support for Numeric Parameters

### LINKED ISSUE #315: Add Numeric Parameter Type
### Problem

Right now, all variables are just simple Strings. It would be nice to have different types. We can add a numeric variable type.

### Solution

The UI should render an input field which only accepts numbers instead of a String field. The back-end should also not accept anything but a number.

### Documentation Changes

- [x] parameter type in `template.conf` is no longer optional, otherwise we cannot parse the default parameter properly ([change](https://github.com/FRosner/cluster-broccoli/wiki/Upgrading-to-0.8.0#parameters-and-types))
- [x] we need a conversion script to fill in a configurable default type for all instances saved previously ([change](https://github.com/FRosner/cluster-broccoli/wiki/Upgrading-to-0.8.0#parameters-and-types))
- [x] explain that there are now `integer` and `decimal` parameters ([change](https://github.com/FRosner/cluster-broccoli/wiki/Templates#parameter-types))

### Links

- #71 (booleans)

sohaibiftikhar:
Fixes #315 

Moved to #360 so that tests can pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 target/
-.idea/
 
 # Artifacts from building docker distributions
 /docker/test/cluster-broccoli-dist/
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 target/
+.idea/
 
 # Artifacts from building docker distributions
 /docker/test/cluster-broccoli-dist/
--- a/server/src/main/scala/de/frosner/broccoli/controllers/InstanceController.scala
+++ b/server/src/main/scala/de/frosner/broccoli/controllers/InstanceController.scala
@@ -134,7 +134,9 @@ object InstanceController {
         .right[InstanceError, Account](loggedIn)
         .ensure(InstanceError.RolesRequired(Role.Administrator): InstanceError)(_.role == Role.Administrator)
       _ <- Either
-        .fromOption(Try(instanceCreation.parameters("id").as[String]).toOption, InstanceError.IdMissing: InstanceError)
+        .fromOption(instanceCreation.parameters.get("id"), InstanceError.IdMissing: InstanceError)
+        .map(idJson => Either.fromOption(idJson.validate[String].asOpt, InstanceError.IdMalformed: InstanceError))
+        .joinRight
         .ensureOr(InstanceError.UserRegexDenied(_, user.instanceRegex): InstanceError)(_.matches(user.instanceRegex))
       newInstance <- Either
         .fromTry(instanceService.addInstance(instanceCreation))
--- a/server/src/main/scala/de/frosner/broccoli/controllers/InstanceController.scala
+++ b/server/src/main/scala/de/frosner/broccoli/controllers/InstanceController.scala
@@ -23,6 +23,8 @@ import play.api.mvc.{Action, Controller, Results}
 import shapeless.tag
 import squants.information.Information
 
+import scala.util.{Failure, Success, Try}
+
 case class InstanceController @Inject()(
     instances: NomadInstances,
     instanceService: InstanceService,
@@ -132,7 +134,7 @@ object InstanceController {
         .right[InstanceError, Account](loggedIn)
         .ensure(InstanceError.RolesRequired(Role.Administrator): InstanceError)(_.role == Role.Administrator)
       _ <- Either
-        .fromOption(instanceCreation.parameters.get("id"), InstanceError.IdMissing: InstanceError)
+        .fromOption(Try(instanceCreation.parameters("id").as[String]).toOption, InstanceError.IdMissing: InstanceError)
         .ensureOr(InstanceError.UserRegexDenied(_, user.instanceRegex): InstanceError)(_.matches(user.instanceRegex))
       newInstance <- Either
         .fromTry(instanceService.addInstance(instanceCreation))
--- a/server/src/main/scala/de/frosner/broccoli/models/Instance.scala
+++ b/server/src/main/scala/de/frosner/broccoli/models/Instance.scala
@@ -2,6 +2,7 @@ package de.frosner.broccoli.models
 
 import de.frosner.broccoli.RemoveSecrets
 import play.api.libs.json._
+import play.api.libs.json.JsNull
 
 import scala.util.{Failure, Success, Try}
 
@@ -55,10 +56,11 @@ object Instance {
         val jsValueMap =
           instance.parameterValues.map {
             case (paramName, paramValue) =>
-              (paramName, paramValue.asJsValue)
+              // TODO: Shouldn't use nulls, Fix this after fixing instanceRemoveSecrets
+              (paramName, Option(paramValue).map(_.asJsValue).getOrElse(JsNull))
           }
         Json.obj(
-          "id"-> instance.id,
+          "id" -> instance.id,
           "template" -> instance.template,
           "parameterValues" -> jsValueMap
         )
@@ -73,10 +75,11 @@ object Instance {
         val jsValueMap =
           instance.parameterValues.map {
             case (paramName, paramValue) =>
-              (paramName, paramValue.asJsValue)
+              // TODO: Shouldn't use nulls, Fix this after fixing instanceRemoveSecrets
+              (paramName, Option(paramValue).map(_.asJsValue).getOrElse(JsNull))
           }
         Json.obj(
-          "id"-> instance.id,
+          "id" -> instance.id,
           "template" -> instance.template,
           "parameterValues" -> jsValueMap
         )
@@ -87,24 +90,34 @@ object Instance {
   implicit val instancePersistenceReads: Reads[Instance] = {
     import Template.templatePersistenceReads
     new Reads[Instance] {
-      override def reads(json: JsValue): JsResult[Instance] = {
+      override def reads(json: JsValue): JsResult[Instance] =
         Try {
           val id: String = (json \ "id").as[String]
           val template: Template = (json \ "template").as[Template]
           val parameterValues: Map[String, ParameterValue] =
-            (json \ "parameterValues").as[JsObject].value.map {
-              case (paramName, paramJsValue) =>
-                ParameterValue.constructParameterValueFromJson(paramName, template, paramJsValue) match {
-                  case Success(paramValue) => (paramName, paramValue)
-                  case Failure(ex) => throw ex
-                }
-            }.toMap // We need to cast an unmodifiable Map to an immutable one. What is the cost for this?
+            (json \ "parameterValues")
+              .as[JsObject]
+              .value
+              .map {
+
+                case (paramName, paramJsValue) =>
+                  paramJsValue match {
+                    // TODO: Shouldn't use nulls, Fix this after fixing instanceRemoveSecrets
+                    case JsNull =>
+                      (paramName, null)
+                    case _ =>
+                      ParameterValue.constructParameterValueFromJson(paramName, template, paramJsValue) match {
+                        case Success(paramValue) => (paramName, paramValue)
+                        case Failure(ex)         => throw ex
+                      }
+                  }
+              }
+              .toMap // We need to cast an unmodifiable Map to an immutable one. What is the cost for this?
           Instance(id, template, parameterValues)
         } match {
-          case Success(s) => JsSuccess(s)
+          case Success(s)  => JsSuccess(s)
           case Failure(ex) => JsError(ex.getMessage)
         }
-      }
     }
   }
 
--- a/server/src/main/scala/de/frosner/broccoli/models/Instance.scala
+++ b/server/src/main/scala/de/frosner/broccoli/models/Instance.scala
@@ -106,7 +106,7 @@ object Instance {
                     case JsNull =>
                       (paramName, null)
                     case _ =>
-                      ParameterValue.constructParameterValueFromJson(paramName, template, paramJsValue) match {
+                      ParameterValue.fromJsValue(paramName, template.parameterInfos, paramJsValue) match {
                         case Success(paramValue) => (paramName, paramValue)
                         case Failure(ex)         => throw ex
                       }
--- a/server/src/main/scala/de/frosner/broccoli/models/Instance.scala
+++ b/server/src/main/scala/de/frosner/broccoli/models/Instance.scala
@@ -53,10 +53,10 @@ object Instance {
     new Writes[Instance] {
       override def writes(instance: Instance) = {
         val jsValueMap =
-          instance.parameterValues.map(
-            param =>
-              (param._1, param._2.asJsValue)
-          )
+          instance.parameterValues.map {
+            case (paramName, paramValue) =>
+              (paramName, paramValue.asJsValue)
+          }
         Json.obj(
           "id"-> instance.id,
           "template" -> instance.template,
@@ -71,10 +71,10 @@ object Instance {
     new Writes[Instance] {
       override def writes(instance: Instance) = {
         val jsValueMap =
-          instance.parameterValues.map(
-            param =>
-              (param._1, param._2.asJsValue)
-          )
+          instance.parameterValues.map {
+            case (paramName, paramValue) =>
+              (paramName, paramValue.asJsValue)
+          }
         Json.obj(
           "id"-> instance.id,
           "template" -> instance.template,
@@ -92,12 +92,13 @@ object Instance {
           val id: String = (json \ "id").as[String]
           val template: Template = (json \ "template").as[Template]
           val parameterValues: Map[String, ParameterValue] =
-            (json \ "parameterValues").as[JsObject].value.map( param =>
-              ParameterValue.constructParameterValueFromJson(param._1, template, param._2) match {
-                case Success(m) => (param._1, m)
-                case Failure(ex) => throw ex
-              }
-            ).toMap
+            (json \ "parameterValues").as[JsObject].value.map {
+              case (paramName, paramJsValue) =>
+                ParameterValue.constructParameterValueFromJson(paramName, template, paramJsValue) match {
+                  case Success(paramValue) => (paramName, paramValue)
+                  case Failure(ex) => throw ex
+                }
+            }.toMap // We need to cast an unmodifiable Map to an immutable one. What is the cost for this?
           Instance(id, template, parameterValues)
         } match {
           case Success(s) => JsSuccess(s)
--- a/server/src/main/scala/de/frosner/broccoli/models/Instance.scala
+++ b/server/src/main/scala/de/frosner/broccoli/models/Instance.scala
@@ -3,11 +3,11 @@ package de.frosner.broccoli.models
 import de.frosner.broccoli.RemoveSecrets
 import play.api.libs.json._
 
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
 
-case class Instance(id: String, template: Template, parameterValues: Map[String, String]) extends Serializable {
+case class Instance(id: String, template: Template, parameterValues: Map[String, ParameterValue]) extends Serializable {
 
-  def requireParameterValueConsistency(parameterValues: Map[String, String], template: Template) = {
+  def requireParameterValueConsistency(parameterValues: Map[String, ParameterValue], template: Template) = {
     val realParametersWithValues = parameterValues.keySet ++ template.parameterInfos.flatMap {
       case (key, info) => info.default.map(Function.const(key))
     }
@@ -20,7 +20,7 @@ case class Instance(id: String, template: Template, parameterValues: Map[String,
 
   requireParameterValueConsistency(parameterValues, template)
 
-  def updateParameterValues(newParameterValues: Map[String, String]): Try[Instance] =
+  def updateParameterValues(newParameterValues: Map[String, ParameterValue]): Try[Instance] =
     Try {
       requireParameterValueConsistency(newParameterValues, template)
       require(newParameterValues("id") == parameterValues("id"), s"The parameter value 'id' must not be changed.")
@@ -32,7 +32,7 @@ case class Instance(id: String, template: Template, parameterValues: Map[String,
       )
     }
 
-  def updateTemplate(newTemplate: Template, newParameterValues: Map[String, String]): Try[Instance] =
+  def updateTemplate(newTemplate: Template, newParameterValues: Map[String, ParameterValue]): Try[Instance] =
     Try {
       requireParameterValueConsistency(newParameterValues, newTemplate)
       require(newParameterValues("id") == parameterValues("id"), s"The parameter value 'id' must not be changed.")
@@ -47,19 +47,64 @@ case class Instance(id: String, template: Template, parameterValues: Map[String,
 }
 
 object Instance {
+
   implicit val instanceApiWrites: Writes[Instance] = {
     import Template.templateApiWrites
-    Json.writes[Instance]
+    new Writes[Instance] {
+      override def writes(instance: Instance) = {
+        val jsValueMap =
+          instance.parameterValues.map(
+            param =>
+              (param._1, param._2.asJsValue)
+          )
+        Json.obj(
+          "id"-> instance.id,
+          "template" -> instance.template,
+          "parameterValues" -> jsValueMap
+        )
+      }
+    }
   }
 
   implicit val instancePersistenceWrites: Writes[Instance] = {
     import Template.templatePersistenceWrites
-    Json.writes[Instance]
+    new Writes[Instance] {
+      override def writes(instance: Instance) = {
+        val jsValueMap =
+          instance.parameterValues.map(
+            param =>
+              (param._1, param._2.asJsValue)
+          )
+        Json.obj(
+          "id"-> instance.id,
+          "template" -> instance.template,
+          "parameterValues" -> jsValueMap
+        )
+      }
+    }
   }
 
   implicit val instancePersistenceReads: Reads[Instance] = {
     import Template.templatePersistenceReads
-    Json.reads[Instance]
+    new Reads[Instance] {
+      override def reads(json: JsValue): JsResult[Instance] = {
+        Try {
+          val id: String = (json \ "id").as[String]
+          val template: Template = (json \ "template").as[Template]
+          val parameterValues: Map[String, ParameterValue] =
+            (json \ "parameterValues").as[JsObject].value.map( param =>
+              ParameterValue.constructParameterValueFromJson(param._1, template, param._2) match {
+                case Success(m) => (param._1, m)
+                case Failure(ex) => throw ex
+              }
+            ).toMap
+          Instance(id, template, parameterValues)
+        } match {
+          case Success(s) => JsSuccess(s)
+          case Failure(ex) => JsError(ex.getMessage)
+        }
+      }
+    }
   }
 
   /**
--- a/server/src/main/scala/de/frosner/broccoli/models/ParameterType.scala
+++ b/server/src/main/scala/de/frosner/broccoli/models/ParameterType.scala
@@ -7,7 +7,6 @@ import scala.collection.immutable
 sealed trait ParameterType extends EnumEntry with EnumEntry.Lowercase
 
 object ParameterType extends Enum[ParameterType] with PlayJsonEnum[ParameterType] {
-  val a: Boolean = true
   val values: immutable.IndexedSeq[ParameterType] = findValues
   case object Raw extends ParameterType
   case object String extends ParameterType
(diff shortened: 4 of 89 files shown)


.gitignore
```diff
@@ -1,4 +1,5 @@
 target/
+.idea/
```
FRosner: Please include .idea to your [global ignore file](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) as otherwise we end up with ignores for every possible editor out there :)

server/src/main/scala/de/frosner/broccoli/templates/TemplateRenderer.scala
```diff
@@ -7,12 +7,25 @@ import de.frosner.broccoli.models.{Instance, ParameterInfo, ParameterType}
...
+
 /**
   * Renders json representation of the passed instance
   *
   * @param defaultType The default type of template parameters
   */
 class TemplateRenderer(defaultType: ParameterType) {
+
+  /** Checks if the value is a number or a is a number **/
+  def isNumber(value: String): Boolean = {
```
FRosner: I think the safest and best way to check if something can be a number is to try to convert it. Everything else usually has corner cases you didn't think about.

IMO we should check what datatype the JSON number from play expects (I believe it might be BigDecimal or Double), try to parse the string in a `Try` and return that.

webui/src/Views/ParameterFormView.elm
```diff
@@ -416,73 +454,112 @@ newParameterValueView template parameterInfos maybeInstanceParameterForm enabled
...
+
+getErrorMsg: String -> DataType -> Maybe String
+getErrorMsg value dataType =
+    case dataType of
+        -- we do not validate string and raw fields
+        StringParam -> Nothing
+        RawParam -> Nothing
+        NumericParam ->
+            let
+                expr = regex "^-?[0-9]+(\\.[0-9]+)?$"
```
FRosner: Also it sounds like a bad idea to check the param again. I think we should store the parameters with correct types and then simply decode them.

As Scala doesn't like Maps with multiple types we can either

- store the type information somewhere and have dynamic casting (not so optimal), or
- have one map per parameter type which is type safe and insert them in a multi-branch when injecting the template configuration. I.e. we have a `rawParameters` map, a `stringParameters` map, and a `numericParameters` map? To avoid too many values in the case class we can always have a wrapper case class grouping all the parameter values / infos. Not sure if this one is trivial to implement, also because we map the template.conf structure 1 to 1 to our case class, but I prefer something type safe and sound over storing strings and validating / parsing them on the fly. This gets especially tricky when you think about other parameter types like sets, etc. How are you going to represent them safely as strings and parse? The logic will be very error prone.

FRosner (changes requested):

webui/src/Views/ParameterFormView.elm
FRosner: Ah I see there is the need to also encode the parameters in the front-end. Sorry for my confusion. Then you actually need two validation steps, or? But the back-end should rely on JSON serialization and the front-end should send appropriate types over, i.e. a number as a number and not quoted as a string.

You can use toInt and toFloat http://package.elm-lang.org/packages/elm-lang/core/latest/String

FRosner:
Feel free to ping me and we can have a quick call / webex to discuss the design of the solution.

server/src/main/scala/de/frosner/broccoli/templates/TemplateRenderer.scala
sohaibiftikhar: Had a discussion with Aleks and I am moving away from regex.

webui/src/Views/ParameterFormView.elm
sohaibiftikhar: We also discussed this. It seems a bit involved right now to send stuff as specific types instead of strings (because of the Map stuff that you said). Also, JSON may not be enough for some future types we may need to support for example positive integers (for CPU/memory etc.) or ranged integers (for ports) and so on. I would suggest we take that up as a separate issue maybe.

FRosner: It seems involved but storing everything as strings in a huge map and then trying to convert it with custom logic seems wrong. It might look nice and "clean" for numbers but even simple structures like lists require you to duplicate basic JSON/HOCON serialization logic. Also we should fail fast when loading the templates and not at the point where we try to use them.

I also think that the argument for positive integers is not a very strong one. Custom types holding positive integers are usually a wrapper for normal integers and not strings. IMO we should make use of the existing types as much as possible and add additional stuff on top as required.

Here's what I think could be a good solution:

Take

```
parameters = {
  "id" = {
    type = string
  }
  "json" = {
    type = raw
  }
  "cpu" = {
    type = number
  }
}
```

and parse it to [`TemplateConfig.TemplateInfo`](https://github.com/FRosner/cluster-broccoli/blob/master/server/src/main/scala/de/frosner/broccoli/templates/TemplateConfig.scala) as it is already done. In [`ParameterInfo. fromTemplateInfoParameter `](https://github.com/FRosner/cluster-broccoli/blob/master/server/src/main/scala/de/frosner/broccoli/models/ParameterInfo.scala#L20) we add the logic to build up the individual properly typed parameters, also validating the default parameters to fail fast. 

Having one map type for each parameter type sounds like additional work, but it's a simple solution IMO. However, it will blow up when we add more and more parameter types so maybe there is a smarter solution. We can have `Map[String, (Any, ParameterType)]` and then cast it according to the type before using it.

The front-end code then needs to send the appropriate types. If we have something like a positive int, we specify custom JSON encoders and decoders.

@Gerrrr what do you think?

sohaibiftikhar: > Having one map type for each parameter type sounds like additional work, but it's a simple solution IMO. However, it will blow up when we add more and more parameter types so maybe there is a smarter solution. We can have Map[String, (Any, ParameterType)] and then cast it according to the type before using it.

I am somehow averse to casts (also will probably need to do it repeatedly everywhere we need it). I would rather move towards an inheritance chain of some sort maybe. But I can't visualize this very precisely right now.

FRosner: Inheritance chain sounds scary :D

Gerrrr: >As Scala doesn't like Maps with multiple types we can either

We could define an enum with supported types and say that the map's values are from this enum. Every case from the enum would have its sanitization logic.

FRosner: That's what I was suggesting. But the compiler wants a type for the map. And this would be any then. Sounds like a use case for union types to me but I don't know if we want to go down that road... http://milessabin.com/blog/2011/06/09/scala-union-types-curry-howard/ :D

Gerrrr: > As Scala doesn't like Maps with multiple types we can either

Or have a trait for a template parameter type and classes extending this trait. Then, the map would be of type e.g. `[String, TemplateParameterType]`.

FRosner: You can always use pattern matching to match the types.

```scala
val parameterValues: Map[String, Any] = ???
parameterValues.mapValues {
  case s: String => ???
  case n: BigDecimal => ???
  case _ => ???
}
```

If we want to limit the type scope, having a generic parameter value container like it's done with `JsValue` is another way.

Something like `ParameterValue` as the super type and then subtypes like `StringParameterValue`, `RawParameterValue`. Then pattern matching is much nicer.

FRosner: > Or have a trait for a template parameter type and classes extending this trait. Then, the map would be of type e.g. [String, TemplateParameterType].

It can't be a map to the type. It needs to be a map to the value. But otherwise I agree. The map to the type is given by the parameter infos.

FRosner: So it would be `Map[String, ParameterValue]` and then we have different `ParameterValue` implementations which provide JSON/HOCON serdes accordingly and this gives us validation out of the box.

sohaibiftikhar: > Or have a trait for a template parameter type and classes extending this trait. Then, the map would be of type e.g. [String, TemplateParameterType].

This was what I was talking about exactly (maybe I used the wrong word but it is still inheritance isn't it? :D)

> It can't be a map to the type. It needs to be a map to the value. But otherwise I agree. The map to the type is given by the parameter infos.

And yes this. In the end, we end up with a type map and a value map. I was thinking maybe we could combine this (since you have the types anyway with the trait stuff). This is the part I said I couldn't visualise.

FRosner: I think it doesn't make sense to combine it from a conceptual point of view as the parameter infos are available after template parsing but the values can be changed during operation time by user requests.

I agree with having two maps as discussed above. Ping me if you want to discuss another WIP or final version.

I think we don't need a call now.

sohaibiftikhar:
@FRosner @Gerrrr Could you please have a look now and see if this is at least moving in the right direction? Functions to see specifically: 
1. `InstanceService::addInstance` and 
2. `InstanceService::updateInstance`

server/src/main/scala/de/frosner/broccoli/models/Instance.scala
```diff
@@ -47,19 +47,64 @@ case class Instance(id: String, template: Template, parameterValues: Map[String,
...
 object Instance {
+
   implicit val instanceApiWrites: Writes[Instance] = {
     import Template.templateApiWrites
-    Json.writes[Instance]
+    new Writes[Instance] {
+      override def writes(instance: Instance) = {
+        val jsValueMap =
+          instance.parameterValues.map(
+            param =>
```
FRosner: Please pattern match:

```scala
instance.parameterValues.map {
  case (..., ...) => (..., ...)
}
```

server/src/main/scala/de/frosner/broccoli/models/Instance.scala
```diff
@@ -47,19 +47,64 @@ case class Instance(id: String, template: Template, parameterValues: Map[String,
...
 
   implicit val instancePersistenceWrites: Writes[Instance] = {
     import Template.templatePersistenceWrites
-    Json.writes[Instance]
+    new Writes[Instance] {
+      override def writes(instance: Instance) = {
+        val jsValueMap =
+          instance.parameterValues.map(
+            param =>
+              (param._1, param._2.asJsValue)
```
FRosner: See above

server/src/main/scala/de/frosner/broccoli/models/Instance.scala
```diff
@@ -47,19 +47,64 @@ case class Instance(id: String, template: Template, parameterValues: Map[String,
...
   implicit val instancePersistenceReads: Reads[Instance] = {
     import Template.templatePersistenceReads
-    Json.reads[Instance]
+    new Reads[Instance] {
+      override def reads(json: JsValue): JsResult[Instance] = {
+        Try {
+          val id: String = (json \ "id").as[String]
+          val template: Template = (json \ "template").as[Template]
+          val parameterValues: Map[String, ParameterValue] =
+            (json \ "parameterValues").as[JsObject].value.map( param =>
```
FRosner: see above

server/src/main/scala/de/frosner/broccoli/models/ParameterValue.scala
```diff
@@ -0,0 +1,82 @@
...
+                case ParameterType.Decimal =>
+                    DecimalParameterValue(jsValue.as[BigDecimal])
+            }
+        } match {
+            case Success(s) => Some(s)
+            case Failure(_) => None
+        }
+    }
+}
+trait ParameterValue {
```
FRosner: I suggest to use a sealed trait here.

server/src/main/scala/de/frosner/broccoli/models/ParameterValue.scala
```diff
@@ -0,0 +1,82 @@
...
+    def isEmpty: Boolean
```
FRosner: What is the `isEmpty` method used for?

server/src/main/scala/de/frosner/broccoli/services/InstanceService.scala
```diff
@@ -218,13 +217,27 @@ class InstanceService @Inject()(nomadClient: NomadClient,
...
           potentialTemplate
             .map { template =>
-              val maybeNewInstance = Try(Instance(id, template, parameters))
-              maybeNewInstance.flatMap { newInstance =>
-                val result = instanceStorage.writeInstance(newInstance)
-                result.foreach { writtenInstance =>
-                  instancesMap = instancesMap.updated(id, writtenInstance)
+              Try{
+                instanceCreation.parameters.map { param =>
+                  ParameterValue.constructParameterValueFromJson(param._1, template, param._2) match {
```
FRosner: pattern match

server/src/main/scala/de/frosner/broccoli/services/InstanceService.scala
```diff
@@ -236,99 +249,115 @@ class InstanceService @Inject()(nomadClient: NomadClient,
 
   def updateInstance(id: String,
                      statusUpdater: Option[JobStatus],
-                     parameterValuesUpdater: Option[Map[String, String]],
+                     parameterValuesUpdater: Option[Map[String, JsValue]],
```
FRosner, line 251: We use `Map[String, JsValue]` in some places and `Map[String, ParameterValue]` in others. Why is that?

server/src/main/scala/de/frosner/broccoli/templates/TemplateRenderer.scala
```diff
@@ -7,12 +7,26 @@ import de.frosner.broccoli.models.{Instance, ParameterInfo, ParameterType}
...
+
 /**
   * Renders json representation of the passed instance
   *
   * @param defaultType The default type of template parameters
   */
 class TemplateRenderer(defaultType: ParameterType) {
+
+  // Checks if the value is a number
+  private def isNumber(value: String): Boolean = {
```
FRosner: Is this still used?

server/src/main/scala/de/frosner/broccoli/templates/TemplateRenderer.scala
```diff
@@ -22,6 +36,20 @@ class TemplateRenderer(defaultType: ParameterType) {
       case ParameterType.Raw => value
       case ParameterType.String =>
         StringEscapeUtils.escapeJson(value)
+      case ParameterType.Integer =>
```
FRosner: Shouldn't sanitization now work depending on the value? Why is the value string? It should be a parameter value, or?

FRosner (commented):
I started reviewing but I didn't finish, yet. I'm going to continue tomorrow. I didn't look at the front-end code, yet and I didn't browse the back-end code in detail.

The general direction seems to be the right one 👍

server/src/main/scala/de/frosner/broccoli/templates/TemplateRenderer.scala
sohaibiftikhar: Nope. Will be cleaned up.

server/src/main/scala/de/frosner/broccoli/templates/TemplateRenderer.scala
sohaibiftikhar: Yes, this function will be removed. Since the tests used it I left it for now. But will clean everything up tomorrow.

server/src/main/scala/de/frosner/broccoli/models/ParameterValue.scala
sohaibiftikhar: Not sure. Used to remove params with zero length strings from the paramvalues map before. Maybe won't be required anymore after frontend changes.

server/src/main/scala/de/frosner/broccoli/services/InstanceService.scala
sohaibiftikhar: We use JsValue until the point where we can actually decode it. (see `InstanceService::updateInstance`). Once we have the information (from the template.parameterinfo) we can parse it accordingly.

sohaibiftikhar:
Thanks for the review. Frontend stuff is still incomplete so you can postpone the detailed review for later. Just wanted to know the general direction for now.

server/src/main/scala/de/frosner/broccoli/controllers/InstanceController.scala
```diff
@@ -132,7 +134,7 @@ object InstanceController {
         .right[InstanceError, Account](loggedIn)
         .ensure(InstanceError.RolesRequired(Role.Administrator): InstanceError)(_.role == Role.Administrator)
       _ <- Either
-        .fromOption(instanceCreation.parameters.get("id"), InstanceError.IdMissing: InstanceError)
```
FRosner: There should be `EitherT.fromTry` or?

server/src/main/scala/de/frosner/broccoli/models/Instance.scala
```diff
@@ -47,19 +48,77 @@ case class Instance(id: String, template: Template, parameterValues: Map[String,
...
+                    case JsNull =>
+                      (paramName, null)
+                    case _ =>
+                      ParameterValue.constructParameterValueFromJson(paramName, template, paramJsValue) match {
+                        case Success(paramValue) => (paramName, paramValue)
+                        case Failure(ex)         => throw ex
+                      }
+                  }
+              }
+              .toMap // We need to cast an unmodifiable Map to an immutable one. What is the cost for this?
```
FRosner, line 115: I guess it copies the map.

server/src/main/scala/de/frosner/broccoli/models/ParameterType.scala
```diff
@@ -10,4 +10,6 @@ object ParameterType extends Enum[ParameterType] with PlayJsonEnum[ParameterType
   val values: immutable.IndexedSeq[ParameterType] = findValues
   case object Raw extends ParameterType
   case object String extends ParameterType
+  case object Integer extends ParameterType
+  case object Decimal extends ParameterType
```
FRosner, line 14: Why do we differentiate between integer and decimal? In the end what gets serialized is anyway going to be a JS number. What is the type backing up `JsNumber` from play? `BigDecimal`? `Double`? Should we use that one as the numeric type?

server/src/main/scala/de/frosner/broccoli/models/ParameterValue.scala
```diff
@@ -0,0 +1,92 @@
...
+
+import scala.util.{Failure, Success, Try}
+
+object ParameterValue {
+
+  implicit val parameterValueWrites: Writes[ParameterValue] = new Writes[ParameterValue] {
+    override def writes(paramValue: ParameterValue) = paramValue.asJsValue
+  }
+
+  def constructParameterValueFromTypesafeConfig(parameterType: ParameterType,
```
FRosner: Maybe we can shorten the method name to `fromConfigValue`?

server/src/main/scala/de/frosner/broccoli/models/ParameterValue.scala
```diff
@@ -0,0 +1,92 @@
...
+        case ParameterType.String =>
+          StringParameterValue(configValue.unwrapped.asInstanceOf[String])
+        case ParameterType.Integer =>
+          IntParameterValue(configValue.unwrapped().asInstanceOf[Int])
+        case ParameterType.Decimal =>
+          DecimalParameterValue(configValue.unwrapped().asInstanceOf[BigDecimal])
+      }
+    }
+
+  def constructParameterValueFromJson(parameterName: String,
```
FRosner: `fromJsValue`?

server/src/main/scala/de/frosner/broccoli/models/ParameterValue.scala
```diff
@@ -0,0 +1,92 @@
...
+                                      template: Template,
```
FRosner: Why do we need the whole template here? I guess the parameter infos should be enough, or?

The `ParameterValueParsingException` doesn't really need the template ID. It should contain the parameter name and the template abstraction level should be with the caller IMO.

server/src/main/scala/de/frosner/broccoli/models/ParameterValue.scala
```diff
@@ -0,0 +1,92 @@
...
+      val parameterInfo =
+        template.parameterInfos.getOrElse(parameterName, ParameterInfo(parameterName, None, None, None, None, None))
+      constructParameterValueFromJson(parameterInfo.`type`.getOrElse(ParameterType.Raw), jsValue) match {
+        case Some(param) => param
+        case None =>
+          throw ParameterValueParsingException(template.id, parameterName)
+      }
+    }
+
+  def constructParameterValueFromJson(parameterType: ParameterType, jsValue: JsValue): Option[ParameterValue] =
```
FRosner: `fromJsValue`?

server/src/main/scala/de/frosner/broccoli/models/ParameterValue.scala
```diff
@@ -0,0 +1,92 @@
...
+        case ParameterType.Decimal =>
+          DecimalParameterValue(jsValue.as[BigDecimal])
+      }
+    } match {
+      case Success(s) => Some(s)
+      case Failure(_) => None
+    }
+}
+sealed trait ParameterValue {
+  def isEmpty: Boolean
```
FRosner: Why do we need `isEmpty`?

server/src/main/scala/de/frosner/broccoli/services/InstanceService.scala
```diff
@@ -218,13 +217,29 @@ class InstanceService @Inject()(nomadClient: NomadClient,
...
+                    case Success(s)  => (param._1, s)
+                    case Failure(ex) => throw ex
+                  }
                 }
-                result
+              } match {
+                case Success(params) =>
+                  // Remove params with default values
+                  val parameters = params.filter { param =>
+                    !param._2.isEmpty
```
FRosner: Please pattern match :)

server/src/main/scala/de/frosner/broccoli/services/InstanceService.scala
```diff
@@ -236,99 +251,116 @@ class InstanceService @Inject()(nomadClient: NomadClient,
...
-            if (parameterValuesUpdatesWithPossibleDefaults.isDefined) {
-              // New parameter values are specified
-              val newParameterValues = parameterValuesUpdatesWithPossibleDefaults.get
-              instance.updateTemplate(template, newParameterValues)
-            } else {
-              // Just use the old parameter values
-              instance.updateTemplate(template, instance.parameterValues)
+      Try {
+        parameterValuesUpdater.map { p =>
+          p.map { e =>
```
FRosner: Please pattern match :)

server/src/main/scala/de/frosner/broccoli/templates/TemplateConfig.scala
```diff
@@ -1,13 +1,55 @@
...
+import de.frosner.broccoli.models.{ParameterType, ParameterValue}
+import pureconfig.ConfigReader
+import pureconfig.error.{ConfigReaderFailures, ThrowableFailure}
+
+import scala.collection.JavaConversions._
+import scala.util.{Failure, Success, Try}
 
 object TemplateConfig {
 
+  implicit val configReader: ConfigReader[TemplateInfo] = new ConfigReader[TemplateInfo] {
```
FRosner, line 13: Wow why did you define your own config reader? Could it be better to keep the default behaviour from pure-config and parse the default value as `String` like it was. We then can parse the default value into the real type using the type information? Just thinking that this looks hard to maintain.

FRosner (commented):
I think we're on the right track. What do you think @Gerrrr / @aurelianrl ?

server/src/main/scala/de/frosner/broccoli/models/ParameterType.scala
sohaibiftikhar: If you have things like CPU then decimals do not make sense. There is no 1 to one mapping between JSON types and the types we define here. This was the idea. But I can combine them if you think it makes more sense.

server/src/main/scala/de/frosner/broccoli/models/ParameterValue.scala
sohaibiftikhar: Because we filter on empty parameters in the [`addInstance`](https://github.com/FRosner/cluster-broccoli/blob/master/server/src/main/scala/de/frosner/broccoli/services/InstanceService.scala#L209) and [`updateInstance`](https://github.com/FRosner/cluster-broccoli/blob/master/server/src/main/scala/de/frosner/broccoli/services/InstanceService.scala#L245) methods in the `InstanceService`. Since this only made sense for `String` params to me I defaulted the others to true.

server/src/main/scala/de/frosner/broccoli/services/InstanceService.scala
sohaibiftikhar: Missed this somehow.

server/src/main/scala/de/frosner/broccoli/templates/TemplateConfig.scala
sohaibiftikhar: IMHO if we go the String way it kind of lands us into the same problem as before. Maybe @Gerrrr can comment.

server/src/main/scala/de/frosner/broccoli/controllers/InstanceController.scala
sohaibiftikhar: I used `toOption` on the Try object but you are right I am squeezing two different errors into a single one. Will split it into two. Also, `JsValue` is not validated properly (`as[String]` does not guarantee it was a `JsString`)

sohaibiftikhar:
@FRosner Please review again. Also, it would be great if you can have a look at the [integration test logs](https://travis-ci.org/sohaibiftikhar/cluster-broccoli/jobs/365706281) and explain the failures.
Update: I linked my own travis above. It has some weird errors about brocolli not starting. I think its better if I use the one at [frosner](https://travis-ci.org/FRosner/cluster-broccoli/jobs/365706303). Will figure out the failures from there. So you can skip that now.

sohaibiftikhar:
@FRosner The current versioning for a template depends on `Map.toString`. Since this does not look deterministic I changed it to an ordered sequence. As a result, I had to modify the version number in the tests. I believe this would not affect saved instances since the version is not persisted?

server/src/main/scala/de/frosner/broccoli/models/ParameterType.scala
FRosner: I think we shouldn't overcomplicate things. If Nomad fails when specifying 200.4 MHz CPU then we should have ints to fail fast. If Nomad doesn't care, I'd prefer to merge them especially as in JSON it doesn't make a difference.

What do you think @Gerrrr, @aurelianrl ?

server/src/main/scala/de/frosner/broccoli/models/ParameterValue.scala
FRosner: I think we should refactor this. If we require this check maybe we can change the semantics of adding instances. If you want we can have a quick call to discuss that.

FRosner:
> @FRosner Please review again. Also, it would be great if you can have a look at the integration test logs and explain the failures. Update: I linked my own travis above. It has some weird errors about brocolli not starting. I think its better if I use the one at frosner. Will figure out the failures from there. So you can skip that now.

I will add you as a collaborator and then you can create a local branch which will use the travis.

FRosner:
> @FRosner The current versioning for a template depends on Map.toString. Since this does not look deterministic I changed it to an ordered sequence. As a result, I had to modify the version number in the tests. I believe this would not affect saved instances since the version is not persisted?

It will. But it's not too big of a deal. People can safely upgrade their templates. I remember we had a discussion with @Gerrrr  about this versioning some time ago. Where do you convert it to an ordered sequence (at which point in the code)?

The PR is so big that I prefer if we review it in a pair session (webex or in person) and finalize it. Otherwise it will ping back and forth and I have to start understanding everything again. What do you think?

FRosner:
[test.json.txt](https://github.com/FRosner/cluster-broccoli/files/1914638/test.json.txt)

Please check that loading this JSON from the instances directory works when we require the default parameter.

sohaibiftikhar:
@FRosner The script that @Gerrrr wrote is very similar to what I need for transforming the existing instances so that all parameters have a `type`. Do we combine them to a single script or keep them separate as `instances-0.7.0-to-0.8.0-1.py` and `instances-0.7.0-to-0.8.0-2.py`?

PR #361, opened 2018-04-19
Show parameter types in front-end

### LINKED ISSUE #365: Show parameter types in UI
See #361

FRosner:
I think it's good to somehow show the parameter type in the front-end. Otherwise the user has to guess whether he has to enter a number or a string, e.g.

I'm not sure about the visuals. It looks a bit clumsy like this. @sohaibiftikhar what do you think? What do the others think?

image

image

---

fixes #365

### DIFF
--- a/webui/src/Views/ParameterFormView.elm
+++ b/webui/src/Views/ParameterFormView.elm
@@ -16,7 +16,6 @@ import Set exposing (Set)
 import Maybe
 import Maybe.Extra exposing (isJust, join)
 import Json.Decode
-import Regex exposing (contains, regex)
 import Tuple exposing (first, second)
 
 
--- a/webui/src/Views/ParameterFormView.elm
+++ b/webui/src/Views/ParameterFormView.elm
@@ -340,12 +340,9 @@ editParameterValueView instance parameterValues parameterInfos maybeInstancePara
                                             [ ( "background-color", Maybe.withDefault normalParamColor (Maybe.map (\v -> editingParamColor) maybeEditedValue) )
                                             ]
                                         ]
-                                        [ text parameterName ]
-                                  , span
-                                        [ class "input-group-addon" ]
-                                        [ icon
-                                            (dataTypeToIcon dataType)
-                                            [ title (dataTypeToTitle dataType) ]
+                                        [ text parameterName
+                                        , text " "
+                                        , sup [] [ text (dataTypeToTitle dataType) ]
                                         ]
                                   , input
                                         [ type_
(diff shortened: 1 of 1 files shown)


sohaibiftikhar:
Seems okay. But what happens when we start using more complex types? Say for example Regex/Lists/Sets/Ports etc. Finding icons might get harder. Maybe just the tooltip is enough? Since we'd have more flexibility with strings.

webui/src/Views/ParameterFormView.elm
```diff
@@ -196,12 +195,12 @@ templateSelectionView currentTemplate selectedTemplate templates instance maybeR
                   ]
                 ]
             )
-            (List.append
-                [ templateOption currentTemplate selectedTemplate currentTemplate ]
-                (templatesWithoutCurrentTemplate
+            (List.concat
```
sohaibiftikhar, line 198: Good catch. Thanks.

webui/src/Views/ParameterFormView.elm
```diff
@@ -703,3 +716,35 @@ getErrorMessage maybeValue paramName dataType =
...
+            "glyphicon glyphicon-text-size"
+
+        RawParam ->
+            "glyphicon glyphicon-asterisk"
+
+        IntParam ->
+            "fa fa-hashtag"
+
+        DecimalParam ->
+            "glyphicon glyphicon-sound-5-1"
```
sohaibiftikhar: This kind of proves my point. Already scratching around for icons 😀

FRosner:
> Finding icons might get harder. Maybe just the tooltip is enough? Since we'd have more flexibility with strings.

I agree that it might get harder to find icons. On the other hand icons are more intuitive and smaller. If there is no icon available we might as well design one. A tooltip alone is not so nice as it takes some time to load, is not obvious, and not available on mobile.

Despite the fact that we need to have icons for all the upcoming types, what do you say about the ones I chose and the way it is displayed?

sohaibiftikhar:
> If there is no icon available we might as well design one

Well if we do that we may need to design it for all types otherwise the design won't be consistent across types. Generally speaking I would prefer to avoid this. 

About intuitiveness we are basically looking at icons thinking this image may represent type T. Looking at a hash for example a user could conclude that it is a number but does not really know that the number must be an integer (so there is still doubt). And for the decimal the '5.1' sign may be standardized for something else (sound 5.1). For someone who already knows what it is generally used for, it might even become confusing to see it here. What I mean to say is that the icons are not a value add in every case. They work for simple forms but the forms in Broccoli are a bit more 'technical'. Plus it is hard to maintain because of the designing thing.

As for the design itself I think it is good. Does not change too much of what was being done earlier and even though it occupies a tad more space than I would prefer it is still acceptable.

FRosner:
@sohaibiftikhar can you please review again?

sohaibiftikhar:
LGTM

sohaibiftikhar (approved):

